PR #9, opened 2017-09-27
Read from db

atodd-noms:


server/controllers/dashboardController.js
```diff
@@ -1,27 +1,80 @@
...
+function getOffenderNomisIds(releases) {
+    return releases.map(offender => offender.nomisId);
+}
+
+function parseDashboardInfo(upcomingReleases, activeLicences) {
+
+    const required = upcomingReleases.map(offender => {
+        const licence = activeLicences.find(licence => licence.nomisId === offender.nomisId);
+        return licence ? Object.assign(offender, {inProgress: true, licenceId: licence.id}) : offender;
+    });
```
whitfield-mj: With this version of node we can use the spread operator to merge I believe. 

Bit less verbose: {...obj1, ...obj2}

PR #53, opened 2017-10-31
Enforce capitalisation

atodd-noms:


server/routes/prisonerDetails.js
```diff
@@ -21,7 +21,8 @@ module.exports = function({logger, prisonerDetailsService, licenceService, authe
 
         const details = {
             prisonerInfo,
-            moment: require('moment')
+            moment: require('moment'),
+            setCase: require('case')
         };
```
whitfield-mj: Fine for now. I think we should look at a better way to set case and date formats. Passing to the view doesn't feel right. Some sort of pre-processing

PR #55, opened 2017-10-31
Send page

atodd-noms:


server/data/licenceClient.js
```diff
@@ -66,5 +66,19 @@ module.exports = {
 
             getCollection(sql, null, resolve, reject);
         });
+    },
+
+    updateStatus: function(nomisId, status) {
+        return new Promise((resolve, reject) => {
+            const sql = `UPDATE LICENCES SET STATUS = @status 
+                         WHERE NOMIS_ID = @nomisId`;
```
whitfield-mj: I would suggest using normal ' rather than string literal ` for instances that don't require string literal features. Prevents the formatting of the text being retained that can cause problems in tests.

PR #57, opened 2017-11-02
Confirm send

atodd-noms:
Not sure about adding establishment address lookup in licence service, but doesn't seem to warrant a whole new "service" on its own


server/views/confirmation/index.pug
```diff
@@ -0,0 +1,17 @@
+extends ../layout
+
+block content
+    p
```
whitfield-mj: Is that p supposed to be there?

server/data/establishmentsClient.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const apiUrl = config.establishments.apiUrl;
+
+module.exports = {
+    findById: function(agencyId) {
+        const path = `${apiUrl}/establishments/${agencyId}`;
+        return doGet(path);
+        }
+    };
```
whitfield-mj: formatting error, extra tabs before }s

PR #77, opened 2017-11-22
prepopulate additional conditions form if submission already exists i…

whitfield-mj:
…n the database

### DIFF
--- a/server/routes/additionalConditions.js
+++ b/server/routes/additionalConditions.js
@@ -15,7 +15,9 @@ module.exports = function({logger, conditionsService, licenceService}) {
         logger.debug('GET /additionalConditions');
 
         const nomisId = req.params.nomisId;
-        const conditions = await conditionsService.getAdditionalConditions();
+        const existingLicence = await licenceService.getLicence(req.params.nomisId);
+        const licence = existingLicence ? existingLicence.licence : null;
+        const conditions = await conditionsService.getAdditionalConditions(licence);
 
         return res.render('additionalConditions/index', {nomisId, conditions});
     }));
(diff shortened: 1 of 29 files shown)


server/routes/additionalConditions.js
```diff
@@ -15,7 +15,9 @@ module.exports = function({logger, conditionsService, licenceService}) {
         logger.debug('GET /additionalConditions');
 
         const nomisId = req.params.nomisId;
-        const conditions = await conditionsService.getAdditionalConditions();
+        const existingLicence = await licenceService.getLicence(req.params.nomisId);
```
romidane: Only problem here is that your not handling errors that would be thrown from 
`await licenceService.getLicence(req.params.nomisId);`
and 
`await conditionsService.getAdditionalConditions(licence);`

It's ugly but you might want to wrap this code in a try catch :/

server/services/conditionsService.js
```diff
@@ -8,12 +8,18 @@ module.exports = function createConditionsService(licenceClient) {
         }
     }
 
-    async function getAdditionalConditions() {
+    async function getAdditionalConditions(licence = null) {
         try {
             const conditions = await licenceClient.getAdditionalConditions();
 
+            if (licence && licence.additionalConditions) {
```
romidane: Using ramda `path` would remove the need of a default value `licence = null` as well as simplify your if statment to `if (additionalConditions) ...`
But that's a just another way of writting thid

server/services/licenceService.js
```diff
@@ -8,7 +8,7 @@ const {
 
 module.exports = function createLicenceService(licenceClient, establishmentsClient) {
 
-    async function getLicence(nomisId) {
+    async function getLicence(nomisId, {populateConditions = false} = {}) {
```
romidane: what's the value of defaulting `populateConditions` to false.. if it doesn't exist it will be a falsy.. I think you can remove the default value here?

PR #78, opened 2017-11-22
Merge pull request #77 from noms-digital-studio/pre-populate-conditions

whitfield-mj:
prepopulate additional conditions form if submission already exists i…

### DIFF
--- a/server/routes/dischargeAddress.js
+++ b/server/routes/dischargeAddress.js
@@ -16,16 +16,16 @@ module.exports = function({logger, dischargeAddressService, licenceService, auth
         logger.debug('GET /dischargeAddress');
 
         const nomisId = req.params.nomisId;
-        const licence = await licenceService.getLicence(nomisId);
+        const rawLicence = await licenceService.getLicence(nomisId);
+        const licence = rawLicence && rawLicence.licence || null;
 
-        if(licence.length < 1) {
+        if(!licence) {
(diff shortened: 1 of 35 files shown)


server/routes/dischargeAddress.js
```diff
@@ -16,16 +16,16 @@ module.exports = function({logger, dischargeAddressService, licenceService, auth
         logger.debug('GET /dischargeAddress');
 
         const nomisId = req.params.nomisId;
-        const licence = await licenceService.getLicence(nomisId);
+        const rawLicence = await licenceService.getLicence(nomisId);
```
romidane: Youre not handling the error from the promise `licenceService.getLicence(nomisId);`
Simplest solution is to wrap this in a try catch

server/routes/dischargeAddress.js
```diff
@@ -16,16 +16,16 @@ module.exports = function({logger, dischargeAddressService, licenceService, auth
...
+        const licence = rawLicence && rawLicence.licence || null;
```
romidane: I'm a big fan of using http://ramdajs.com/docs/#path
It would simply this to 
`const licence = path(['licence'], rawLicence)`;

You could also just import only that module.. navigating dynamic objects in a safe way it a win IMO

whitfield-mj: Agreed

PR #79, opened 2017-11-23
Pre populate reporting instructions

whitfield-mj:


server/routes/dischargeAddress.js
```diff
@@ -16,16 +17,16 @@ module.exports = function({logger, dischargeAddressService, licenceService, auth
         logger.debug('GET /dischargeAddress');
 
         const nomisId = req.params.nomisId;
-        const licence = await licenceService.getLicence(nomisId);
+        const rawLicence = await licenceService.getLicence(nomisId);
+        const licence = getIn(rawLicence, ['licence']);
 
-        if(licence.length < 1) {
+        if(!licence) {
```
romidane: now this one is a controversial one so feel free to ignore it if you want... but i'm not a big fan of the `!` it's too easy to get missed.. now for the controversial part.. ramda as a utility function called `not` which is reminisent to pyhton or haskell... you could also easly write your own... 

`if(!licence) {`

would become

`if(not(licence)) {`

Again ignore this if you want... just been a JS pendantic about it ahah

server/utils/functionalHelpers.js
```diff
@@ -0,0 +1,11 @@
+const R = require('ramda');
+
+module.exports = {
+    getIn
+};
+
+// pass in your object and a path in array format
+// http://ramdajs.com/docs/#path
+function getIn(object, pathArray) {
```
romidane: I'm happy with hidding underlying libraries 👍

test/routes/reportingInstructionsTest.js
```diff
@@ -26,12 +26,42 @@ const app = appSetup(createReportingInstructionRoute({
...
             .expect('Content-Type', /html/);
     });
+
+    it('calls getLicence from licenceService', () => {
+        return request(app)
+            .get('/1')
+            .expect(200)
+            .expect('Content-Type', /html/)
+            .expect(res => {
+                expect(licenceServiceStub.getLicence).to.be.calledOnce();
```
romidane: What's the final result of going to this url? like what would the user see.. I'm not a big fan of peaking in to the implementation of controller if it can be helped.. I'd be tempted to cover this in an integration test?

server/routes/dischargeAddress.js
whitfield-mj: I have heard this opinion before. Not sure that I agree though! 

Surely it is a part of the language that has been around forever and having to import something to do the same thing is over the top. Also missing it would completely change the functionality of the function and would be caught by any test

PR #95, opened 2018-01-10
retrieve and display image for tasklist page

whitfield-mj:


server/data/nomisClientBuilder.js
```diff
@@ -58,6 +58,22 @@ module.exports = function(token) {
...
+        },
+
+        getImageData: async function(id) {
+            try {
+                const result = await superagent
+                    .get(`${apiUrl}/images/${id}/data`)
+                    .set('Authorization', `Bearer ${generateApiGatewayToken()}`)
+                    .set('Elite-Authorization', token)
+                    .responseType('blob');
+
```
atodd-noms: could call nomisGet with an extra param for responseType

whitfield-mj: Agreed.

PR #96, opened 2018-01-10
Eligibility check page

atodd-noms:


server/utils/licenceFactory.js
```diff
@@ -50,6 +51,23 @@ function createConditionsObject(selectedConditions, formInputs) {
     }, {});
 }
 
+function createEligibilityObject(object) {
+    const acceptedAttributes = Object.keys(licenceModel.eligibility);
+
+    const filterIfUnwanted = {
```
whitfield-mj: This name confused me a bit, sounds like a method. What about something like requiredSelectionForAttribute or something similar. Not easy to name!

server/utils/licenceFactory.js
```diff
@@ -50,6 +51,23 @@ function createConditionsObject(selectedConditions, formInputs) {
...
+        excludedReasons: 'excluded',
+        unsuitableReasons: 'unsuitable'
+    };
+
+    const filteredAcceptedAttributes = acceptedAttributes.filter(attribute => {
+        const decider = filterIfUnwanted[attribute];
```
whitfield-mj: Suggest: const requiredSelectorChecked = requiredSelectionForAttribute[attribute]. Wordy but I find it tricky to follow currently

PR #102, opened 2018-01-15
added started status to caselist if licence exists

whitfield-mj:


server/services/caseListService.js
```diff
@@ -1,8 +1,8 @@
 const logger = require('../../log.js');
-const {isEmpty} = require('../utils/functionalHelpers');
+const {isEmpty, getIn} = require('../utils/functionalHelpers');
```
romidane: `getInThere` 😂

PR #104, opened 2018-01-18
Location string

atodd-noms:


server/services/utils/formatForView.js
```diff
@@ -1,12 +1,16 @@
 const moment = require('moment');
 const setCase = require('case');
-module.exports = {formatObjectForView};
+module.exports = {formatObjectForView, formatObjectForViewWithOptions};
 
-function formatObjectForView(object, options) {
+function formatObjectForView(object) {
+    return formatObjectForViewWithOptions(object, {capitalise: nameFields, dates: dateFields, custom: customFields});
+}
```
whitfield-mj: Not sure there is any value in a function that just calls another one. Would have been tempted to just have one function as all we need passing in is the object itself (as options don't seem to change at this point)

server/services/utils/formatForView.js
```diff
@@ -1,12 +1,16 @@
...
+
+function formatObjectForViewWithOptions(object, options) {
     return Object.keys(object).reduce((builtObject, itemKey) => {
 
         if (object[itemKey] instanceof Object && !Array.isArray(object[itemKey])) {
-            return {...builtObject, [itemKey]: formatObjectForView(object[itemKey], options)};
+            return {...builtObject, [itemKey]: formatObjectForViewWithOptions(object[itemKey], options)};
         }
 
         if (options.dates && options.dates.includes(itemKey)) {
```
whitfield-mj: don't need to check if options.dates exists now as it is hard coded

server/services/utils/formatForView.js
```diff
@@ -17,6 +21,57 @@ function formatObjectForView(object, options) {
             return {...builtObject, [itemKey]: setCase.capital(object[itemKey].toLowerCase())};
         }
 
+        if (options.custom && options.custom[itemKey]) {
```
whitfield-mj: don't need to check if options. custom exists now as it is hard coded

server/services/utils/formatForView.js
```diff
@@ -17,6 +21,57 @@ function formatObjectForView(object, options) {
...
+    return offences && offences[0] ? offences[0].offenceDescription : '';
+}
+
+function formatCom(com) {
+    const name = com && com[0] ? [com[0].firstName, com[0].lastName].join(' ') : '';
+    return setCase.capital(name.toLowerCase());
+}
+
+function formatAliases(aliasesList) {
+    return aliasesList && aliasesList[0] ?aliasesList.map(alias => {
```
whitfield-mj: probably want space after ?

PR #111, opened 2018-01-24
192 BASS referral form

atodd-noms:


server/routes/address.js
```diff
@@ -0,0 +1,47 @@
...
+    }));
+
+    return router;
+};
+
+async function updateFromSubmission(nomisId, body, licenceService) {
+    const existingLicence = await licenceService.getLicence(nomisId);
+    if (!existingLicence) throw new Error('Bass referral without a licence');
+
+    await licenceService.updateBassReferral(body);
```
romidane: Does this function just fire of an not return anything i.e. a successful update?

server/services/licenceService.js
```diff
@@ -106,6 +107,18 @@ module.exports = function createLicenceService(licenceClient, establishmentsClie
         }
     }
 
+    async function updateBassReferral(data = {}) {
+        try {
+            const nomisId = data.nomisId;
+            const bassReferralData = createBassReferralObject(data);
+
+            return await licenceClient.updateSection('bassReferral', nomisId, bassReferralData);
```
romidane: you actually don't need to do
`return await licenceClient.updateSection('bassReferral', nomisId, bassReferralData);`
You can just do
`return licenceClient.updateSection('bassReferral', nomisId, bassReferralData);`

Because the calling function is the one what will be awaiting the response. `licenceClient.updateSection` returns a promise so does `await licenceClient.updateSection`

server/services/licenceService.js
```diff
@@ -106,6 +107,18 @@ module.exports = function createLicenceService(licenceClient, establishmentsClie
...
+        } catch (error) {
+            console.error('Error during setBassReferral', error.stack);
+            throw error;
+        }
+    }
+
     async function sendToOmu(nomisId) {
         try {
             return await licenceClient.updateStatus(nomisId, 'SENT');
```
romidane: while your here since you can go ahead and remove async from this function since it's not needed

server/views/address/bassReferral.pug
```diff
@@ -0,0 +1,35 @@
+extends ../layout
+
+block content
+
+    - var bassReferralValue = bassReferral ? bassReferral.bassReferralRequested : null;
```
romidane: can this not be returned from the router?

server/views/address/bassReferralForm.pug
```diff
@@ -0,0 +1,12 @@
+block bassReferralForm
+
+    - var vals = bassReferral ? bassReferral : '';
```
romidane: same here?

test/services/licenceServiceTest.js
```diff
@@ -184,7 +203,7 @@ describe('licenceService', () => {
             );
         });
 
-        it('should merge eligibility object with exosting data', () => {
+        it('should merge eligibility object with existing data', () => {
```
romidane: 👍

server/services/licenceService.js
atodd-noms: I deleted all this stuff, not sure why it's back in

PR #114, opened 2018-01-26
Lic 192 process bass request

whitfield-mj:


whitfield-mj:
Have implemented a configurable routing system so that we only need one GET and POST per route file. Also wrote an update function that should be usable under all circumstances and not need the factory work

PR #115, opened 2018-01-29
Proposed address

whitfield-mj:


server/views/proposedAddress/includes/otherResidents.pug
```diff
@@ -0,0 +1,18 @@
+- var residentsToShow = data.residents &&  data.residents.length > 3 ? data.residents.length : 3
```
romidane: Maybe not now but we will have to think of a way to keep logic out of the templates

PR #120, opened 2018-02-01
added address submission page. Added ability to update status

whitfield-mj:


server/routes/config/proposedAddress.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
             fieldToDecideOn: 'decision',
             Yes: '/hdc/taskList/',
             No: '/hdc/proposedAddress/bassReferral/'
-        }
+        },
+        nextPath: null
```
romidane: why we setting these to null now?

PR #125, opened 2018-02-06
Stop passing role manually

atodd-noms:
In a couple of places we were passing role to the view but it's already there.

I've changed it from `profile.roleCode` to `user.role` so in any view you can always do user.role

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -151,7 +151,7 @@ module.exports = function createApp({
     app.locals.asset_path = '/public/';
 
     function addTemplateVariables(req, res, next) {
-        res.locals.profile = req.user;
+        res.locals.user = req.user;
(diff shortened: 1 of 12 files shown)


server/routes/licenceDetails.js
```diff
@@ -9,7 +9,7 @@ module.exports = function({logger, licenceService, authenticationMiddleware}) {
         logger.debug('GET /licenceDetails');
         const licence = await licenceService.getLicence(req.params.nomisId, {populateConditions: true});
 
-        res.render(`licenceDetails/${req.user.roleCode}`, licence);
+        res.render('licenceDetails/', licence);
```
romidane: trailing slash?

PR #126, opened 2018-02-06
added conditions pages to licenceConditions journey

whitfield-mj:
Sorry is quite large as a lot of moving things and deleting things


server/routes/config/licenceConditions.js
```diff
@@ -1,4 +1,15 @@
 module.exports = {
+    standardConditions: {
+        nextPathDecision: {
+            fieldToDecideOn: 'additionalConditions',
```
atodd-noms: fieldToDecideOn -> discriminator?

server/routes/config/licenceConditions.js
```diff
@@ -1,4 +1,15 @@
...
+            Yes: '/hdc/licenceConditions/additionalConditions/',
+            No: '/hdc/licenceConditions/riskManagement/'
+        }
+    },
+    conditionsReview: {
+        licenceSection: 'additionalConditions',
+        nextPath: '/hdc/licenceConditions/riskManagement/'
+    },
     riskManagement: {
```
atodd-noms: I'm not convinced about having risk management as part of licence conditions but it seems I'm behind the times re state-of-the-art HDC routing logic magic

whitfield-mj: Probably agree. The original idea was to grop them all together in the 5 journeys. Will be worth revisiting afterwards I think.

PR #135, opened 2018-02-08
Send sent

atodd-noms:
Suggestion for consideration.
Some limitations because we don't yet restrict seeing/editing licences after it's been handed over to the other user.


server/routes/send.js
```diff
@@ -13,29 +14,21 @@ module.exports = function({logger, licenceService, authenticationMiddleware}) {
...
 
-        res.render('send', {nomisId: req.params.nomisId});
-    }));
-
-    router.post('/omu/', asyncMiddleware(async (req, res) => {
-        logger.debug('POST /send/omu');
-
-        const nomisId = req.body.nomisId;
+        const {nomisId} = req.params;
+        const licence = await licenceService.getLicence(nomisId, {populateConditions: true});
```
whitfield-mj: Probably don't need to populate conditions do you?

atodd-noms: ta. Copy error.

PR #141, opened 2018-02-12
LIC-219: Connect routes

romidane:
In this PR, I add test and refactor the routes config.

As an aside I've added default times for the curfew hours based on the designs

### DIFF
--- a/server/utils/routes.js
+++ b/server/utils/routes.js
@@ -2,18 +2,42 @@ module.exports = {
     getPathFor
 };
 
-
-function decidePath({decisionInfo, data}) {
-    const decidingValue = data[decisionInfo.discriminator];
-    return decisionInfo[decidingValue];
+function getPathFor({data, config}) {
+    return decidePath({nextPath: config.nextPath, data});
 }
 
-function getPathFor({formName, data, formConfig}) {
-    if (formConfig[formName].nextPath) {
-        return formConfig[formName].nextPath;
+function decidePath({nextPath, data}) {
+    if (!nextPath.decisions) {
+        return nextPath.path;
     }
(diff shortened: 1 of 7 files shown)


server/utils/routes.js
```diff
@@ -2,18 +2,40 @@ module.exports = {
     getPathFor
 };
```
whitfield-mj: This page seems like it is upside down to me. Would be easier to read starting with getPathFor and then reading downwards

server/utils/routes.js
```diff
@@ -2,18 +2,40 @@ module.exports = {
...
-    if (formConfig[formName].nextPathDecision) {
-        return decidePath({decisionInfo: formConfig[formName].nextPathDecision, data});
+
+    if (nextPath.decisions) {
+        return getPathFromAnswer({nextPath: nextPath.decisions, data}) || nextPath.path;
     }
-    return null;
+
+
+    return nextPath.path;
```
whitfield-mj: This can be tidied by initially checking !nextPath.decisions and returning nextPath.path. Then you don't need to keep checking for nextPath.decisions

test/utils/routesTest.js
```diff
@@ -0,0 +1,115 @@
...
+                       {
+                           discriminator: 'barAnswer',
+                           No: '/baz'
+                       },
+                       {
+                           discriminator: 'bazAnswer',
+                           No: '/foo'
+                       }
+                   ],
+                   path: '/foo'
```
whitfield-mj: Not a good test as the path that it should turn into is the same as the fall-through test, you don't know which one you are getting. You need a different value for 'path' as for bazAnswer's 'No'

PR #142, opened 2018-02-13
added ability to save one bespoke conditions

whitfield-mj:


server/routes/hdc.js
```diff
@@ -51,17 +52,19 @@ module.exports = function({logger, licenceService, conditionsService, authentica
...
                 nomisId,
                 conditions,
+                bespokeConditions,
                 submissionError: true
             });
         }
 
-        await licenceService.updateLicenceConditions(validatedInput);
+        await licenceService.updateLicenceConditions(nomisId, additionalConditions, bespokeConditions);
         return res.redirect('/hdc/licenceConditions/conditionsSummary/' + nomisId);
```
romidane: you don't need the return.. I know it's not your change but while your there

PR #146, opened 2018-02-13
Passed full licence into curfew address review

whitfield-mj:


server/views/licenceConditions/curfewAddressReviewForm.pug
```diff
@@ -2,16 +2,20 @@ extends ../layout
 
 block content
 
+    -var review = (data.licenceConditions) ? data.licenceConditions.curfewAddressReview : null
```
romidane: can you just make that an empty object?

PR #158, opened 2018-02-19
2 licence status calculator

atodd-noms:


server/utils/licenceStatus.js
```diff
@@ -0,0 +1,324 @@
...
+        case licenceStages.ELIGIBILITY:
+            return [getEligibilityStageState(data)];
+        default:
+            return [];
+    }
+}
+
+const combiner = (acc, data) => {
+    Object.assign(acc.tasks, {...data.tasks});
+    Object.assign(acc.decisions, {...data.decisions});
```
whitfield-mj: {...data.tasks} is equivalent to data.tasks

whitfield-mj: Do you think this is more readable? Seems more in-keeping with what we do elsewhere
`const combiner = (acc, data) => {
    const combinedTasks = {...acc.tasks, ...data.tasks};
    const combinedDecisions = {...acc.decisions, ...data.decisions};

    return {
        ...acc,
        tasks: combinedTasks,
        decisions: combinedDecisions
    };
};
`

server/utils/licenceStatus.js
```diff
@@ -0,0 +1,324 @@
...
+            return [getEligibilityStageState(data), getRoStageState(data), getApprovalStageState(data)];
+        case licenceStages.PROCESSING_CA:
+        // fall through
+        case licenceStages.PROCESSING_RO:
+            return [getEligibilityStageState(data), getRoStageState(data)];
+        case licenceStages.ELIGIBILITY:
+            return [getEligibilityStageState(data)];
+        default:
+            return [];
+    }
```
whitfield-mj: Don't object to this method. If it were me, I'd do something like: 
```
function getRequiredState(stage, data) {

    const config = {
        [licenceStages.DECIDED]: [getEligibilityStageState, getRoStageState, getApprovalStageState],
        [licenceStages.APPROVAL]: [getEligibilityStageState, getRoStageState, getApprovalStageState],
        [licenceStages.PROCESSING_CA]: [getEligibilityStageState, getRoStageState],
        [licenceStages.PROCESSING_RO]: [getEligibilityStageState, getRoStageState],
        [licenceStages.ELIGIBILITY]: [getEligibilityStageState],
    };

    return config[stage].map(getStateMethod => getStateMethod(data));
}

```

atodd-noms: Yeah that's nicer

server/utils/licenceStatus.js
atodd-noms: I had something like that a one point. Have applied this change

PR #159, opened 2018-02-19
3 label calculator

atodd-noms:


server/utils/licenceStatusLabels.js
```diff
@@ -0,0 +1,124 @@
+const {licenceStages} = require('../models/licenceStages');
+const {taskStates} = require('../models/taskStates');
+
+module.exports = {getStatusLabel};
+
+
+function getStatusLabel(licenceStatus, role) {
+
+    if(!licenceStatus || !licenceStatus.stage || !licenceStatus.decisions || !licenceStatus.tasks) {
+        return 'Not started';
```
atodd-noms: Will re-use the 'allEmpty' function we just talked about

whitfield-mj: could use your new allEmpty function

PR #163, opened 2018-02-20
6 use licence status in tasklist

atodd-noms:


server/utils/licenceStatus.js
```diff
@@ -6,7 +6,7 @@ module.exports = {getLicenceStatus};
 
 function getLicenceStatus(licenceRecord) {
 
-    if (allEmpty([licenceRecord, licenceRecord.licence, licenceRecord.status])) {
+    if (!licenceRecord || !licenceRecord.licence || !licenceRecord.status) {
```
whitfield-mj: Why did you decide against allEmpty?

whitfield-mj: I suppose allEmpty is incorrect isn't it, you want 'anyEmpty'

atodd-noms: if licenceRecord is null, licenceRecord.licence throws an error....

PR #196, opened 2018-03-06
Added recursion to allow control of nested form elements

whitfield-mj:


whitfield-mj:
Previously nested form elements (such as those used in the proposed address form) would just be added to licence. This will give us control of nested fields so that, e.g. we can put validation tags on them in the config file

PR #226, opened 2018-04-06
move addresses to be stored in an array

whitfield-mj:


server/views/proposedAddress/curfewAddress.pug
```diff
@@ -15,7 +15,7 @@ block content
         input(type="hidden" name="nomisId" value=nomisId || '')
         div.borderBottomLight.paddingBottom.largeMarginBottom
             h3.heading-medium Preferred address
-            +addressForm('preferred')
+            +addressForm(0)
```
romidane: can we create temp vars for `0` and one `1` like `preferred = 1`, `alternative = 1`... It's not obvious what each form is unless your paying attention to the text above them

PR #227, opened 2018-04-10
moved address review items into array

whitfield-mj:
Moved the address review results into the array item.

When we know what requirements there are for an address then we could look at using one of the fields as a unique identifier rather than relying on array position.

### DIFF
--- a/server/utils/licenceStatus.js
+++ b/server/utils/licenceStatus.js
@@ -251,24 +251,20 @@ function getCurfewAddressState(licence, optedOut, bassReferralNeeded) {
 
     function getState(licence) {
 
+        const addresses = getIn(licence, ['proposedAddress', 'curfewAddress', 'addresses']);
+
         if (optedOut || bassReferralNeeded) {
             return taskStates.DONE;
         }
 
-        if (isEmpty(getIn(licence, ['proposedAddress', 'curfewAddress']))) {
-            return taskStates.UNSTARTED;
-        }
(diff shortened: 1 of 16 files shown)


server/utils/licenceStatus.js
```diff
@@ -278,24 +274,36 @@ function getCurfewAddressState(licence, optedOut, bassReferralNeeded) {
...
+        return consent || electricity || homeVisitConducted || deemedSafe;
+    });
 
-    if ([consentAnswer, electricityAnswer, deemedSafeAnswer].some(isEmpty)) {
+    if (reviewStarted) {
         return {curfewAddressReview: taskStates.STARTED, curfewAddressApproved: 'unfinished'};
     }
 
-    return {curfewAddressReview: taskStates.DONE, curfewAddressApproved: 'approved'};
+    return {curfewAddressReview: taskStates.UNSTARTED, curfewAddressApproved: 'unfinished'};
```
romidane: I think this should be at the top since all the others depend on the presence of an address

test/routes/curfewTest.js
```diff
@@ -32,46 +32,46 @@ describe('/hdc/curfew', () => {
...
-                nextPath: '/hdc/taskList/1'
-            },
-            {
-                url: '/curfew/curfewHours/1',
-                body: {nomisId: 1},
-                section: 'curfewHours',
-                nextPath: '/hdc/taskList/1'
+                nextPath: '/hdc/curfew/addressSafety/5/1'
             }
+            // {
```
romidane: Comments?

PR #232, opened 2018-04-18
added sorting for caslist

whitfield-mj:


test/services/caseListServiceTest.js
```diff
@@ -136,5 +136,78 @@ describe('caseListService', () => {
...
+                ]);
+
+                const result = await service.getHdcCaseList(user);
+
+                expect(result[0].name).to.eql('address1');
+                expect(result[1].name).to.eql('address2');
+                expect(result[2].name).to.eql('address3');
+            });
+
+            it('should order by second', async() => {
```
atodd-noms: ...by conditionalReleaseDate... ?

test/services/caseListServiceTest.js
```diff
@@ -136,5 +136,78 @@ describe('caseListService', () => {
                 expect(nomisClient.getHdcEligiblePrisoners).not.to.be.calledOnce();
             });
         });
+
+        describe('sorting', () => {
+
+            const address1 = {
+                name: 'address1',
```
atodd-noms: you've got addresses on the brain....

test/services/caseListServiceTest.js
```diff
@@ -136,5 +136,78 @@ describe('caseListService', () => {
...
+
+                const result = await service.getHdcCaseList(user);
+
+                expect(result[0].name).to.eql('address3');
+                expect(result[1].name).to.eql('address4');
+                expect(result[2].name).to.eql('address5');
+            });
+        });
     });
 });
```
atodd-noms: It's possible for CRD to be missing and in that case we show the ARD.
If we're showing ARD in the CRD/ARD column then we should probable sort on ARD

PR #244, opened 2018-04-23
Show refusal reason and prevent changes once decision made

atodd-noms:


server/routes/hdc.js
```diff
@@ -306,9 +306,9 @@ module.exports = function({logger, licenceService, conditionsService, prisonerSe
 
         if (req.body.anchor) {
             res.redirect(`${nextPath}${nomisId}#${req.body.anchor}`);
```
whitfield-mj: Do we need to return this now?

PR #250, opened 2018-04-26
First steps for implemintation of licence validation

whitfield-mj:
First iteration of validation of licence. Will likely need to evolve in terms of what is returned to caller when we know how it will be used.

Maybe incorporate validation config into existing form config if it doesn't make too much of a mess

### DIFF
--- a/server/services/licenceService.js
+++ b/server/services/licenceService.js
@@ -11,10 +11,14 @@ const {
     notAllValuesEmpty,
     allValuesEmpty,
     getFirstArrayItems,
-    replaceArrayItem} = require('../utils/functionalHelpers');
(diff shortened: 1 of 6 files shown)


server/services/licenceService.js
```diff
@@ -272,6 +276,24 @@ module.exports = function createLicenceService(licenceClient) {
...
     }
 
+    function validateLicence(licence, stage) {
+
+        const sections = getLicenceSectionToValidate(stage);
+        const validationErrors = sections.map(validate(licence)).filter(item => item);
+
+        if (isEmpty(validationErrors)) {
+            return null;
+        }
```
atodd-noms: or empty list?

PR #257, opened 2018-04-30
Pg4

atodd-noms:


server/services/caseListService.js
```diff
@@ -77,9 +77,9 @@ function getOffenderIds(releases) {
 
 function getStatus(prisoner, licences, role) {
 
-    const licenceForPrisoner = licences.find(rawLicence => {
-        return prisoner.offenderNo === rawLicence.nomisId;
-    });
+    const licenceForPrisoner = licences ? licences.find(rawLicence => {
+        return prisoner.offenderNo === rawLicence.nomis_id;
+    }) : null;
```
whitfield-mj: Is this now needed because tedious getCollection would return an empty array whereas postgres returns a null? 

Would it be easier to handle this in the dbclient so that we return [] if null comes back from db?

PR #259, opened 2018-05-01
added required input validation to additional conditions

whitfield-mj:
Added validation for required fields of additional conditions. Lead to a lot of old stuff being removed which I think will, in turn, mean that some refactoring of old code is necessary. Didn't want to overfill this merge though


server/utils/licenceFactory.js
```diff
@@ -1,4 +1,5 @@
 const {conditionsOrder} = require('../models/conditions');
+const NO_INPUT_MESSAGE = 'INPUT REQUIRED';
```
whitfield-mj: Short term fix to enable us to display that something is missing on the next page. Will be updated when validation display is decided

server/services/utils/conditionsFormatter.js
```diff
@@ -0,0 +1,40 @@
+const moment = require('moment');
+const DATE_FIELD = 'appointmentDate';
+const ALWAYS_REQUIRED = ['additionalConditions', 'nomisId'];
+
+module.exports = {formatConditionsInput};
+
+function formatConditionsInput(inputObject, selectedConditionsConfig) {
```
whitfield-mj: Now that this file doesn't do any validating I think this can be streamlined at some point.

PR #277, opened 2018-05-10
Remove call to bookings/sentenceDetail - same info in offender-sentences

atodd-noms:


server/data/nomisClientBuilder.js
```diff
@@ -120,3 +110,20 @@ async function nomisGet(path, query, token, headers = {}, responseType = '') {
         throw error;
     }
 }
+
+function addReleaseDates(prisoners) {
+    return prisoners.length > 0 ? prisoners.map(prisoner => addReleaseDate(prisoner)) : prisoners;
```
whitfield-mj: Don't think the ternary is necessary. [].map(....)  would return [] anyway

whitfield-mj: In light have that, I don't think I'd have a separate function, just return prisoners.map(addReleaseDate) from where it is called

PR #291, opened 2018-05-17
Get Licences working with locally running elite2-api

Mjwillis:


Mjwillis:
@atodd-noms This is because the auth header is Authentication for the application, it has no knowledge of the "elite2-authentication" header, this is swapped over by the gateway and therefore when by passing, it doesn't find what its looking for (i.e. the Authentication header is empty (or wrong))

PR #300, opened 2018-05-22
Pdf

atodd-noms:


server/routes/pdf.js
```diff
@@ -0,0 +1,36 @@
...
+        return res.redirect('/hdc/pdf/create/' + nomisId);
+    }));
+
+    router.get('/create/:nomisId', asyncMiddleware(async (req, res) => {
+
+        const {nomisId} = req.params;
+        logger.debug(`GET pdf/create/${nomisId}`);
+        const pdf = await pdfService.generatePdf(templateName, nomisId, req.user.token);
+
+        res.writeHead(200, {'Content-Type': 'application/pdf'});
```
whitfield-mj: You could just do res.type('application/pdf') I think

server/services/pdfService.js
```diff
@@ -0,0 +1,48 @@
...
+        logger.info(`Creating PDF at URI '${pdfGenPath}' for template '${templateName}'`);
+
+        try {
+            const result = await superagent
+                .post(pdfGenPath)
+                .send({
+                    templateName,
+                    values
+                });
+            return new Buffer(result.body);
```
whitfield-mj: new Buffer is depricated. Think we need return Buffer.from(result.body)

PR #309, opened 2018-05-30
Add PSS conditions

atodd-noms:
Part of https://dsdmoj.atlassian.net/browse/LIC-227

### DIFF
--- a/server/utils/licenceFactory.js
+++ b/server/utils/licenceFactory.js
@@ -1,5 +1,6 @@
 const {conditionsOrder} = require('../models/conditions');
-const {getIn} = require('../utils/functionalHelpers');
+const {multiFields} = require('../models/conditions');
+const {getIn, flatten} = require('../utils/functionalHelpers');
 
 module.exports = {
     createLicenceObjectFrom,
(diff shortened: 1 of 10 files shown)


server/utils/licenceFactory.js
```diff
@@ -156,18 +159,23 @@ function injectVariablesIntoViewObject(fieldNames, fieldPositionObject, userInpu
...
+    });
 
-    const variableString = (error, variable) => error ? `[${error}]` : variable;
-    const addressString = variableString(addressError, appointmentAddress);
-    const dateString = variableString(dateError, appointmentDate);
-    const timeString = variableString(timeError, appointmentTime);
-    const string = `${addressString} on ${dateString} at ${timeString}`;
+    const fieldErrors = config.fields
+        .map(fieldName => getIn(userErrors, [fieldName]))
+        .filter(e => e);
```
whitfield-mj: Absolutely fine, just by way of alternative you could avoid a second filter loop by using a reduce

server/utils/licenceFactory.js
```diff
@@ -156,18 +159,23 @@ function injectVariablesIntoViewObject(fieldNames, fieldPositionObject, userInpu
...
 
-    const variableKey = anyErrors ? 'error' : 'variable';
+    const string = flatten(strings
+        .map((string, index) => [string, config.joining[index] || '']))
+        .join('');
+
+    const variableKey = fieldErrors.length > 0 ? 'error' : 'variable';
```
whitfield-mj: Could use isEmpty(fieldErrors) ? ... if it makes it more explicit

server/utils/licenceFactory.js
whitfield-mj: Maybe wouldn't be as readable though

PR #326, opened 2018-06-07
Lic 400 ro oauth

atodd-noms:


server/routes/send.js
```diff
@@ -32,12 +35,12 @@ module.exports = function({logger, licenceService, prisonerService, authenticati
         res.redirect('/hdc/sent/' + nomisId);
     }));
 
-    function getSubmissionTarget(nomisId, stage, {tokenId}) {
+    function getSubmissionTarget(nomisId, stage, {role, tokenId}) {
```
whitfield-mj: The reason that tokenId went into an object was so that we could name it, to describe what it was for, because at point of use it seemed strange to be passing username into the get requests. Not sure if adding role like this recreates that confusion.

I wonder if there is a way of naming that object key that makes it clear why you are passing it in at point of use. Something like
`const prisonerInfo = await prisonerService.getPrisonerDetails(nomisId, {
            refreshContext: req.user.role,
            tokenId: req.user.username
});`

Tricky one

atodd-noms: Interesting. Will have a think. I found tokenId confusing. I thought, why is suddenly called token Id when it's really user name? User name was obvious, cos we look up the token for that user and tokens are inherently user-specific

atodd-noms: The caller knows which user is using it. It shouldn't know anything about tokens that are used for A&A. "get these details and here's the user" versus "get these details and by the way the security mechanism underneath is going to need to look up an oauth token with this ID". Our mate Glen would call that "leakage"

romidane: I think my point around this was around this was that any extra details such as the user and the role should be seen as request configuration rather than plain arugments..
The way I see it in my head is as follows

`getSubmissionTarget ` with this nomisId and stage. And btw use these configuration to make your request.

as opposed to 
`getSubmissionTarget `with this nomisId, stage, and the current user is as follow and his role is such and such

The configuration part would be up to us to configure that to our needs.

I get both side of the arguments.. it does read weird if we are passing in a tokenId where is actually a username. might have to go with @atodd-noms  on this one

atodd-noms: I'm going to move the role stuff down into the signInService and tokenStore. That part shouldn't be leaking out. But there's no getting away from the fact that the current user IS the context for the call, so it makes sense to me to say "for this user". One way or another, what we need to know in the back end is "for which user are we calling the API?", and likewise that API integration layer shouldn't really know anything about how the web app manages users.

PR #331, opened 2018-06-11
LIC-264: Fix eligibility logic

romidane:


server/utils/licenceStatus.js
```diff
@@ -163,6 +163,26 @@ function getEligibilityStageState(licence) {
...
+    if (insufficientTime) {
+        return false;
+    }
+
+    if (unsuitable && !exceptionalCircumstances) {
+        return false;
+    }
+
+    return true;
+}
```
whitfield-mj: Is this properly tested?

romidane: My assumption here is that it's just a longer version of `!(excluded || insufficientTime || unsuitable)`
And if the the tests still pass for those then I guess it is?

romidane: Add some test coverage

PR #338, opened 2018-06-13
LIC-421: Display PCA submit button

romidane:


test/routes/taskListTest.js
```diff
@@ -177,6 +177,35 @@ describe('GET /taskList/:prisonNumber', () => {
...
+                return request(app)
+                    .get('/1233456')
+                    .expect(200)
+                    .expect(res => {
+                        expect(res.text).to.include('/hdc/proposedAddress/optOut/');
+                    });
+            });
+        });
+
         context('when offender is ineligible', () => {
```
whitfield-mj: We will now need tests for when the send button is visible on the tasklist

whitfield-mj: We have 3 possible times that it will be visible:
stage = ELIGIBILITY -> caToRo is allowed transition (normal route)
stage =ELIGIBILITY -> caToDm is allowed transition (new route)
stage =PROCESSING_CA -> caToDm  is allowed transition

server/views/taskList/taskList.pug
```diff
@@ -31,6 +31,13 @@ block content
             if (['UNSTARTED', 'ELIGIBILITY'].includes(licenceStatus.stage))
                 include eligibilityTask
                 include proposedAddressTask
+                if licenceStatus.decisions.eligible
+                    if licenceStatus.decisions.insufficientTime && !licenceStatus.decisions.insufficientTimeContinue
+                        include caSubmitRefusalTask
```
atodd-noms: might be better to introduce a new decision `insufficientTimeStop`, based on the combination of insufficientTime and !insufficientTimeContinue, then this would be
` if insufficientTimeStop include caSubmitRefusalTask`

server/views/taskList/eligibilityTask.pug
```diff
@@ -47,18 +47,30 @@ div.largeMarginBottom
...
-                        div.pure-g.borderTopLight.smallPaddingTopBottom
-                            div.pure-u-3-4.padRight15
-                                | Is there less than 4 weeks to go to the offender's CRD?
-                            if licenceStatus.decisions.excluded || unsuitable
+
+                        if licenceStatus.decisions.excluded || unsuitable
+                            div.pure-g.borderTopLight.smallPaddingTopBottom
+                                div.pure-u-3-4.padRight15
+                                    | Is there less than 4 weeks to go to the offender's CRD?
                                 div#crdTimeAnswer.pure-u-1-4.center.bold N/A
```
atodd-noms: not sure what the intent is above but display text doesn't relate to the if

server/views/taskList/caSubmitRefusalTask.pug
```diff
@@ -0,0 +1,22 @@
+- var canSendToDm = licenceStatus.stage === 'ELIGIBILITY' || allowedTransitions.caToDm
+
```
atodd-noms: i don't think you can reuse allowedTransitions.caToDm, certainly _shouldn't_ reuse it

should be caToDmRefusal or something

server/views/taskList/caSubmitRefusalTask.pug
```diff
@@ -0,0 +1,22 @@
...
+    div.borderBottomLight.largeMarginBottom
+
+    div.pure-g
+        div.pure-u-3-4
+            if canSendToDm
+                | Ready to submit
+            else if licenceStatus.decisions.postponed
+                | Submission unavailable - HDC application postponed
+            else if licenceStatus.decisions.finalChecksRefused
+                | Submission unavailable - HDC refused
```
atodd-noms: final checks can't be relevant to this because we can only be in the ELIGIBILITY stage

server/views/taskList/caSubmitRefusalTask.pug
```diff
@@ -0,0 +1,22 @@
...
+    h2.heading-medium Submit to decision maker
+
+    div.borderBottomLight.largeMarginBottom
+
+    div.pure-g
+        div.pure-u-3-4
+            if canSendToDm
+                | Ready to submit
+            else if licenceStatus.decisions.postponed
+                | Submission unavailable - HDC application postponed
```
atodd-noms: Can it be postponed in the ELIGIBILITY stage?

server/views/send/index.pug
```diff
@@ -11,6 +11,8 @@ block content
         if user.role == 'CA'
             if stage == 'PROCESSING_CA'
                 include CAtoDM.pug
+            else if licenceStatus.decisions.dmRefusal
```
atodd-noms: what's `decisions.dmRefusal`

server/utils/licenceStatus.js
```diff
@@ -194,13 +197,30 @@ function getExclusionTaskState(licence) {
...
-
-    const timeAnswer = getIn(licence, ['eligibility', 'crdTime', 'decision']);
+    const decision = getIn(licence, ['eligibility', 'crdTime', 'decision']);
+    const dmApproval = getIn(licence, ['eligibility', 'crdTime', 'dmApproval']);
 
     return {
-        insufficientTime: timeAnswer === 'Yes',
-        crdTime: timeAnswer ? taskStates.DONE : taskStates.UNSTARTED
+        insufficientTimeContinue: dmApproval === 'Yes',
+        insufficientTime: decision === 'Yes',
```
atodd-noms: combine the above two to create ` insufficientTimeStop`, which would be true if insufficientTime=yes && insufficientTimeContinue == no

server/utils/licenceStatusTransitions.js
```diff
@@ -52,18 +52,23 @@ function canSendDmToCa(licenceStatus) {
...
+    const allTaskComplete = required.every(it => it === taskStates.DONE);
+
+    if (decisions.bassReferralNeeded) {
+        return allTaskComplete && tasks.bassReferral == taskStates.DONE;
+    }
+
+    return allTaskComplete;
 }
 
 function canSendCaToDm(licenceStatus) {
```
atodd-noms: this needs updating to account for insufficientTimeStop

or better,m have a new transition called caToDmRefusal

server/utils/licenceStatus.js
```diff
@@ -130,10 +130,16 @@ function getCaStageState(licence) {
...
+    const {insufficientTime, crdTime, insufficientTimeContinue} = getCrdTimeState(licence);
     const {unsuitable, suitability, exceptionalCircumstances} = getSuitabilityState(licence);
     const eligibility = getEligibilityState(unsuitable, excluded, [exclusion, crdTime, suitability]);
-    const eligible = isEligible({excluded, insufficientTime, unsuitable, exceptionalCircumstances});
+    const eligible = isEligible({
+        excluded,
+        unsuitable,
+        exceptionalCircumstances,
+        insufficientTime,
+        insufficientTimeContinue
```
atodd-noms: insufficientTime shouldn't be part of eligible, because we use 'not eligible' to signal 'stop, no further processing required' (probably need to clean up some names in this aspect later) but for insufficient time it will proceed, either to RO or DM

server/views/taskList/caSubmitRefusalTask.pug
romidane: Don't think so

server/views/taskList/proposedAddressTask.pug
```diff
@@ -1,10 +1,12 @@
-if licenceStatus.tasks.eligibility === 'DONE' && licenceStatus.decisions.eligible
+if licenceStatus.tasks.eligibility === 'DONE' && licenceStatus.decisions.eligible && licenceStatus.decisions.insufficientTimeContinue
     div.largeMarginBottom.borderTop
```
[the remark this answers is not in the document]
romidane: I've moved the logic out the partial

server/utils/licenceStatus.js
```diff
@@ -165,18 +172,19 @@ function getEligibilityStageState(licence) {
...
-    if (insufficientTime) {
+
+    if (unsuitable && !exceptionalCircumstances) {
         return false;
     }
 
-    if (unsuitable && !exceptionalCircumstances) {
+    if (insufficientTimeStop) {
         return false;
     }
```
atodd-noms: simplify to `return !insufficientTimeStop`

server/utils/licenceStatus.js
```diff
@@ -194,13 +202,31 @@ function getExclusionTaskState(licence) {
...
 
     return {
-        insufficientTime: timeAnswer === 'Yes',
-        crdTime: timeAnswer ? taskStates.DONE : taskStates.UNSTARTED
+        insufficientTimeContinue: decision === 'Yes' && dmApproval === 'Yes',
+        insufficientTimeStop: decision === 'Yes' && dmApproval === 'No',
+        insufficientTime: decision === 'Yes',
+        crdTime: getState(licence)
     };
+
```
atodd-noms: This isn't clear, and the way insufficientTimeContinue is used in eligibilityTask isn't consistent.
Would suggest something like
```
    return {
        insufficientTime: decision === 'Yes',
        insufficientTimeApproval: dmApproval === 'Yes',

        insufficientTimeStop: decision === 'Yes' && dmApproval === 'No',
        
        crdTime: getState(licence)
    };
```

insuffientTime and insufficientTimeApproval are used to convey the decisions for the two inividual questions based on their answers, whereas insufficientTimeStop is the overall effect of the two things combined.

In the eligibilityTask, this
```
if licenceStatus.decisions.insufficientTimeContinue
                                            | Yes
                                        else
                                            | No
```
won't show the chosen answer, but instead shows the combination

server/utils/licenceStatusTransitions.js
```diff
@@ -52,25 +53,44 @@ function canSendDmToCa(licenceStatus) {
...
-function canSendCaToDm(licenceStatus) {
+function caToDmRefusal(licenceStatus) {
+    const stage = licenceStatus.stage;
+    const decisions = licenceStatus.decisions;
+
+    if ( stage === 'ELIGIBILITY' && decisions.insufficientTimeStop) {
+        return true;
+    }
+
+    return false;
```
atodd-noms: `return stage === 'ELIGIBILITY' && decisions.insufficientTimeStop;`

server/views/taskList/caSubmitRefusalTask.pug
```diff
@@ -0,0 +1,16 @@
+div.largeMarginBottom.borderTop
+    h2.heading-medium Submit to decision maker
+
```
atodd-noms: should this be the "submit for refusal" text?

server/views/send/index.pug
```diff
@@ -11,6 +11,8 @@ block content
         if user.role == 'CA'
             if stage == 'PROCESSING_CA'
                 include CAtoDM.pug
+            else if licenceStatus.decisions.insufficientTimeStop
+                include CAtoDMRefusal.pug
```
atodd-noms: couldn't this be `if allowedTransitions.caToDmRefusal` then we don't need to pass in licenceStatus?

caToDmRefusal accounts for insufficientTimeStop and also stages (and whatever other rules may come in in future)

Allowed transitions and sending go together more tightly than individual decisions

server/views/taskList/caSubmitTask.pug
```diff
@@ -1,23 +1,23 @@
+- var canSendToDm = licenceStatus.stage === 'PROCESSING_CA' || allowedTransitions.caToDm
+- var canSendToRo = licenceStatus.stage === 'ELIGIBILITY' && allowedTransitions.caToRo
+
```
atodd-noms: this isn't needed - allowed transitions tells you what's allowed

Maybe move the stage check into the allowedtransitions logic to make it a tighter spec, but the view should just use allowedtransitions

server/views/taskList/caSubmitTask.pug
```diff
@@ -1,23 +1,23 @@
+- var canSendToDm = licenceStatus.stage === 'PROCESSING_CA' || allowedTransitions.caToDm
+- var canSendToRo = licenceStatus.stage === 'ELIGIBILITY' && allowedTransitions.caToRo
```
atodd-noms: the 'or' looks odd. Surely can't send to DM at any time during final checks?

server/views/taskList/eligibilityTask.pug
atodd-noms: I think there are faults in this logic but I've done some changes in a different story and there's another story so just ignore it and we'll fix in those other stories

server/views/taskList/taskList.pug
```diff
@@ -28,7 +28,15 @@ block content
         if user.role === 'CA'
             if (['UNSTARTED', 'ELIGIBILITY'].includes(licenceStatus.stage))
                 include eligibilityTask
-                include proposedAddressTask
+
+                if licenceStatus.decisions.eligible
+                    if licenceStatus.tasks.eligibility === 'DONE'
```
atodd-noms: remove blank line
combine two if statements into one
Looks to me like you can move the task === DONE check into the licenceStatus.isEligible, then this line cleans up to
`if licenceStatus.decisions.eligible`

server/views/taskList/taskList.pug
```diff
@@ -28,7 +28,15 @@ block content
...
+                        include proposedAddressTask
+                        include caSubmitTask
+                else
+                    include caSubmitRefusalTask
```
atodd-noms: This means if the offender is ineligible for any reason it gets sent to DM. Is that correct, because it didn't use to work that way (a while since I've worked on that aspect) and it wasn't in the story.

server/views/taskList/caSubmitTask.pug
romidane: Seems that way 👍

PR #349, opened 2018-06-18
Fix broken licence status WRT exceptional circumstances

atodd-noms:


server/utils/licenceStatus.js
```diff
@@ -131,9 +131,9 @@ function getCaStageState(licence) {
 function getEligibilityStageState(licence) {
     const {excluded, exclusion} = getExclusionTaskState(licence);
     const {insufficientTime, crdTime} = getCrdTimeState(licence);
-    const {unsuitable, suitability, exceptionalCircumstances} = getSuitabilityState(licence);
-    const eligibility = getEligibilityState(unsuitable, excluded, [exclusion, crdTime, suitability]);
-    const eligible = isEligible({excluded, insufficientTime, unsuitable, exceptionalCircumstances});
+    const {unsuitable, effectiveUnsuitable, suitability, exceptionalCircumstances} = getSuitabilityState(licence);
```
romidane: I'm not sure what `effectiveUnsuitable` means can you give me some context?

atodd-noms: how about `unsuitableResult` ?

some name needed to convey the overall effective unsuitability state formed by the combination of "unsuitable" and "exceptional circumstances"

they may be "unsuitable" but due to exceptional circumstances the effective "result" of the unsuitable check is that they're _not_ unsuitable.

we need to revisit this stuff one day because the eligibility task display forces us to munge together the "answers" given and the "decisions" those answers form, whereas in other places we only need to convey the effective decision.

server/utils/licenceStatus.js
```diff
@@ -294,6 +302,11 @@ function getRiskManagementState(licence) {
 
 function getApprovalState(licence) {
 
+    const refusalReasons = {
```
romidane: Where is this used in the code base?

PR #355, opened 2018-06-26
LIC-467: Application not displaying override dates

romidane:


server/data/nomisClientBuilder.js
```diff
@@ -146,12 +152,68 @@ function nomisGetBuilder(tokenStore, signInService, username) {
...
+    if (automaticReleaseOverrideDate && automaticReleaseOverrideDate !== invalidDate) {
+        return automaticReleaseOverrideDate;
+    }
+
+    if (automaticReleaseDate && automaticReleaseDate !== invalidDate) {
+        return automaticReleaseDate;
+    }
+
+    return null;
+}
```
atodd-noms: how about something like overrideIfPresent(standard, override){ etc }

server/data/nomisClientBuilder.js
```diff
@@ -146,12 +152,68 @@ function nomisGetBuilder(tokenStore, signInService, username) {
...
+        automaticReleaseOverrideDate
+    } = prisoner.sentenceDetail;
+
+    const ard = effectiveAutomaticReleaseDate({automaticReleaseDate, automaticReleaseOverrideDate});
+
+    return {
+        ...prisoner,
+        sentenceDetail: merge(prisoner.sentenceDetail, {effectiveAutomaticReleaseDate: ard})
+    };
+}
```
atodd-noms: couldn't these also be one general function?

server/data/nomisClientBuilder.js
```diff
@@ -146,12 +152,68 @@ function nomisGetBuilder(tokenStore, signInService, username) {
...
+function addReleaseDate(prisoner) {
+    const {
+        automaticReleaseDate,
+        automaticReleaseOverrideDate,
+        conditionalReleaseDate,
+        conditionalReleaseOverrideDate
+    } = prisoner.sentenceDetail;
+
+    const crd = effectiveConditionalReleaseDate({conditionalReleaseDate, conditionalReleaseOverrideDate});
+    const ard = effectiveAutomaticReleaseDate({automaticReleaseDate, automaticReleaseOverrideDate});
```
atodd-noms: couldn't these be passed in instead of recalculating them

server/data/nomisClientBuilder.js
romidane: I like that idea.. will change

PR #360, opened 2018-07-04
Store CSURF details token in cookies

romidane:


server/app.js
```diff
@@ -194,7 +198,7 @@ module.exports = function createApp({
 
     // CSRF protection
     if (!testMode) {
-        app.use(csurf());
+        app.use(csurf({cookie: true}));
```
whitfield-mj: Seems like we are already using this above

romidane: Good spot!

PR #374, opened 2018-07-11
Audit key events

atodd-noms:
https://dsdmoj.atlassian.net/browse/LIC-85

https://dsdmoj.atlassian.net/wiki/spaces/LIC/pages/893550605/Auditing

### DIFF
--- a/server/routes/hdc.js
+++ b/server/routes/hdc.js
@@ -26,7 +26,9 @@ const formConfig = {
     ...approvalConfig
 };
 
-module.exports = function({logger, licenceService, conditionsService, prisonerService, authenticationMiddleware}) {
+module.exports = function(
+    {logger, licenceService, conditionsService, prisonerService, authenticationMiddleware, audit}) {
+
     const router = express.Router();
     router.use(authenticationMiddleware());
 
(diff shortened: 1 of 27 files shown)


server/routes/hdc.js
```diff
@@ -322,3 +379,15 @@ module.exports = function({logger, licenceService, conditionsService, prisonerSe
...
+function auditableData(input) {
+    return Object.keys(input).reduce((objectBuilt, key) => {
+        if (notAcceptedKeys.includes(key)) {
+            return {...objectBuilt};
+        } else {
+            const value = input[key] || '';
+            return {...objectBuilt, ...{[key]: value}};
+        }
+    }, {});
+}
```
whitfield-mj: There is a nice Ramda function https://ramdajs.com/docs/#omit that could be used to filter an object. Would look like `R.omit(['nomisId', '_csrf', 'anchor'], input)`

server/routes/hdc.js
```diff
@@ -322,3 +379,15 @@ module.exports = function({logger, licenceService, conditionsService, prisonerSe
     return router;
 };
 
+function auditableData(input) {
+    return Object.keys(input).reduce((objectBuilt, key) => {
+        if (notAcceptedKeys.includes(key)) {
+            return {...objectBuilt};
```
whitfield-mj: Equivalent to `return objectBuilt`

PR #398, opened 2018-07-23
simplified markForHandover function

whitfield-mj:


server/models/licenceStages.js
```diff
@@ -10,23 +10,16 @@ const licenceStages = {
...
-    CA: {
-        RO: licenceStages.PROCESSING_RO,
-        DM: licenceStages.APPROVAL
+    caToRo: licenceStages.PROCESSING_RO,
+    caToDm: licenceStages.APPROVAL,
+    roToCa: {
+        'default': licenceStages.PROCESSING_CA,
+        addressRejected: licenceStages.ELIGIBILITY,
+        optedOut: licenceStages.ELIGIBILITY
     },
```
whitfield-mj: This has made me think that these two edge cases should have their own transition

atodd-noms: yes

PR #399, opened 2018-07-23
DM CRD Refusal

atodd-noms:
https://dsdmoj.atlassian.net/browse/LIC-455


server/routes/config/approval.js
```diff
@@ -10,5 +10,16 @@ module.exports = {
         nextPath: {
             path: '/hdc/send/decided/'
         }
+    },
+    crdrefuse: {
```
whitfield-mj: Don't we camel case all the others?

PR #448, opened 2018-08-16
only DMs can post an approval

whitfield-mj:


server/routes/config/approval.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
         ],
         nextPath: {
             path: '/hdc/send/decided/'
+        },
+        authorisation: {
+            authorisedRoles: ['DM'],
+            message: 'HDC can only be approved by a decision maker'
         }
```
atodd-noms: Do we need to be so friendly? Why not just a 403?

server/routes/hdc.js
```diff
@@ -343,7 +343,16 @@ module.exports = function(
...
 
+        const {authorisation} = formConfig[formName];
+        const unauthorisedPost = !isEmpty(authorisation) && !authorisation.authorisedRoles.includes(req.user.role);
+
+        if (unauthorisedPost) {
+            req.flash('authorisation', authorisation.message || 'You are not authorised to submit this form');
+            return res.redirect(`/hdc/${sectionName}/${formName}/${nomisId}`);
+        }
+
         return formPost(req, res, sectionName, formName, nomisId);
```
atodd-noms: Isn't it possible to just prevent GET?

PR #476, opened 2018-08-28
Move audit to middleware

atodd-noms:


server/utils/middleware.js
```diff
@@ -60,3 +62,38 @@ function authorisationMiddleware(req, res, next) {
...
     next();
 }
+
+function auditMiddleware(audit, key) {
+    return async (req, res, next) => {
+
+        const bookingId = req.body.bookingId || req.params.bookingId;
+        const inputs = userInputFrom(req.body);
+
+        const pathSegments = req.path.split('/').filter(s => s && s !== bookingId);
```
whitfield-mj: under what circumstance do you need to check that s exists?

PR #493, opened 2018-09-05
LIC 594 - add contact details for RO users

atodd-noms:


server/routes/admin/admin.js
```diff
@@ -46,24 +46,22 @@ module.exports = function(
 
     router.post('/roUsers/edit/:nomisId', async(async (req, res) => {
```
whitfield-mj: If you're only awaiting inside a try/catch then you won't need the async middleware

server/routes/admin/admin.js
```diff
@@ -87,22 +83,32 @@ module.exports = function(
     }));
 
     router.post('/roUsers/add', async(async (req, res) => {
```
whitfield-mj: Same as above

server/services/admin/userService.js
```diff
@@ -10,17 +11,18 @@ module.exports = function createUserService(userClient) {
             await checkExistingDelius(newDeliusId);
```
whitfield-mj: Is it worth doing this in a Promise.all as you have below?

PR #497, opened 2018-09-05
Get all missing groups not just first one

atodd-noms:


server/routes/pdf.js
```diff
@@ -63,8 +63,8 @@ module.exports = function(
             pdfService.getPdfLicenceData(templateName, bookingId, licence, req.user.token)
         ]);
 
-        const incompleteGroups = Object.keys(missing).find(group => missing[group].mandatory);
-        const canPrint = !incompleteGroups;
+        const incompleteGroups = Object.keys(missing).filter(group => missing[group].mandatory);
+        const canPrint = !(incompleteGroups && incompleteGroups.length > 0);
```
whitfield-mj: might be clearer to use `const canPrint = isEmpty(incompleteGroups)`

PR #501, opened 2018-09-06
opted out cannot be sent to RO. display orange box on task list

whitfield-mj:


server/views/taskList/tasks/caSubmitAddressReviewTask.pug
```diff
@@ -4,12 +4,14 @@ div.largeMarginBottom
...
+            if licenceStatus.decisions.optedOut
+                | Submission unavailable - Offender has opted out of HDC
+            else if ['STARTED', 'DONE'].includes(licenceStatus.tasks.curfewAddress)
                 | Ready to submit
             else
                 | Submission unavailable - Tasks not yet complete
 
         div.pure-u-1.pure-u-md-1-4.center
-            if ['STARTED', 'DONE'].includes(licenceStatus.tasks.curfewAddress)
+            if ['STARTED', 'DONE'].includes(licenceStatus.tasks.curfewAddress) && !licenceStatus.decisions.optedOut
```
atodd-noms: This could be lifted up and done together before the opted-in bits

if opted out
    div.pure-g
        blah offender has opted out
else
    div.pure-g
        all original blah for opted in case

PR #510, opened 2018-09-12
Show RO contact details

atodd-noms:
Can't merge until Nomis does their API update to show delius identifier in COM relation response

see https://dsdmoj.atlassian.net/browse/LIC-603

### DIFF
--- a/server/views/contact/ro.pug
+++ b/server/views/contact/ro.pug
@@ -7,7 +7,7 @@ block content
             include ../includes/back
 
     div.pure-g.largePaddingBottom
-        div.pure-u-1.pure-u-2-3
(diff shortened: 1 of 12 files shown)


server/views/contact/ro.pug
```diff
@@ -0,0 +1,33 @@
+extends ../layout
+
+block content
+
+    div.pure-g
+        div.pure-u-1
+            include ../includes/back
+
+    div.pure-g.largePaddingBottom
+        div.pure-u-1.pure-u-2-3
```
whitfield-mj: This overwrites 100% width column with 66% . Think you need div.pure-u-1.pure-u-md-2-3

PR #511, opened 2018-09-12
Remember form inputs on error. Reuse user details form

atodd-noms:


server/data/userClient.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
 
         const {rows} = await db.query(query);
 
-        return rows;
+        return rows.map(user => convertPropertyNames(user));
```
whitfield-mj: can be written `return rows.map(convertPropertyNames);`

server/data/userClient.js
```diff
@@ -107,7 +107,20 @@ module.exports = {
...
```
whitfield-mj: as above

PR #514, opened 2018-09-13
Alternative caselist

whitfield-mj:


server/services/utils/caseListFormatter.js
```diff
@@ -63,3 +64,45 @@ function compareReleaseDates(address1, address2) {
...
+
+    if (weeks >= 2) {
+        return {text: `${weeks} weeks`};
+    }
+
+    if (Math.sign(days) === 1) {
+        const unit = days === 1 ? 'day' : 'days';
+        return {text: `${days} ${unit}`};
+    }
+
```
atodd-noms: Aren't any of the built in things any use? eg http://momentjs.com/docs/#/displaying/fromnow/

server/services/utils/caseListFormatter.js
```diff
@@ -63,3 +64,45 @@ function compareReleaseDates(address1, address2) {
...
+function addRoleSpecificDecoration(prisoner, role) {
+    if (role === 'CA') {
+        return addHDCEDCountdown(prisoner);
+    }
+    return prisoner;
+}
+
+function addHDCEDCountdown(prisoner) {
+    const dueDate = moment(prisoner.sentenceDetail.homeDetentionCurfewEligibilityDate, 'DD/MM/YYYY');
```
atodd-noms: not sure if sentenceDetail is never null. could use getIn

PR #521, opened 2018-09-14
Clarefication

atodd-noms:
https://dsdmoj.atlassian.net/browse/LIC-546


server/views/caseList/CATable.pug
```diff
@@ -48,7 +48,7 @@ table#hdcEligiblePrisoners.largeMarginBottom
 
                     td
                         if ['UNSTARTED', 'ELIGIBILITY', 'PROCESSING_CA', 'DECIDED', 'MODIFIED', 'MODIFIED_APPROVAL'].includes(prisoner.stage)
-                            if ['Not started', 'Reviewing case'].includes(prisoner.status)
+                            if ['Not started', 'Reviewing case', 'Address rejected'].includes(prisoner.status)
```
whitfield-mj: Just until it is decided to get rid of one, you will need to add this to caseList/alt/CATable too otherwise I suspect this will disappear when the file gets deleted

PR #545, opened 2018-09-28
updated docker-compose to use mock api. Added seed for demo data

whitfield-mj:


package.json
```diff
@@ -23,8 +23,9 @@
     "start-watchmode": "nodemon --watch server server.js",
     "knex": "knex",
     "knex-migrate": "knex-migrate",
-    "migrate": "knex migrate:latest",
-    "seed": "knex seed:run"
+    "migrate": "knex migrate:latest --env standard",
```
atodd-noms: will this work with knex-migrate?

PR #558, opened 2018-10-04
Reorg sending. Have to get target twice to preserve auditing

atodd-noms:


server/routes/send.js
```diff
@@ -20,44 +17,48 @@ module.exports = function({licenceService, prisonerService, authenticationMiddle
...
+        decided: {type: 'dmToCa', receiver: 'CA'},
+        'return': {type: 'dmToCaReturn', receiver: 'CA'},
+        refusal: {type: 'caToDmRefusal', receiver: 'DM'},
+        addressRejected: {type: 'roToCaAddressRejected', receiver: 'CA'},
+        optedOut: {type: 'roToCaOptedOut', receiver: 'CA'}
+    };
+
+    function getSubmissionTarget(target, bookingId, token) {
+        return prisonerService.getOrganisationContactDetails(target, bookingId, token);
+    }
```
whitfield-mj: Doesn't seem to be any point in this function, may as-well just call directly where needed

PR #561, opened 2018-10-08
NN-1200: Split out auth into separate URL

petergphillips:


petergphillips:
Need to amend the tests before this is ready!

PR #566, opened 2018-10-10
Step 1 of all the BASS referral stuff

atodd-noms:
Move BASS referral into its own section (cos more forms coming)
Allow validation and send to RO

### DIFF
--- a/server/routes/bassReferral.js
+++ b/server/routes/bassReferral.js
@@ -1,0 +1,33 @@
+const express = require('express');
+const {async, checkLicenceMiddleWare, authorisationMiddleware, auditMiddleware} = require('../utils/middleware');
+const createStandardRoutes = require('./routeWorkers/standard');
+
+module.exports = function({licenceService, prisonerService, authenticationMiddleware, audit}) {
+
+    const router = express.Router();
+    router.use(authenticationMiddleware());
+    router.param('bookingId', checkLicenceMiddleWare(licenceService, prisonerService));
+    router.param('bookingId', authorisationMiddleware);
+
+    const audited = auditMiddleware(audit, 'UPDATE_SECTION');
+
+    router.use(function(req, res, next) {
+        if (typeof req.csrfToken === 'function') {
+            res.locals.csrfToken = req.csrfToken();
+        }
+        next();
+    });
+
+    const formConfig = require('./config/bassReferral');
+    const standard = createStandardRoutes({formConfig, licenceService, sectionName: 'bassReferral'});
+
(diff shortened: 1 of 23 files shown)


server/services/utils/licenceValidation.js
```diff
@@ -57,6 +57,7 @@ function getMessage(errorType, errorMessage, errorPath) {
         .filter(pathItem => !Number.isInteger(pathItem))
         .join('_');
 
+    // Shouldn't this be first? Custom messages for postcode or telephone don't get used
```
whitfield-mj: Could be. It was originally done this way to give you option for more specific message like 'Enter a date that is not in the past' rather than a catch all message for the field. If we're going to have a defined message for every field then I guess all the stuff above can completely go

server/views/review/includes/bassRequest.pug
```diff
@@ -0,0 +1,15 @@
+include ./itemAndError
+-var request = data.bassReferral && data.bassReferral.bassRequest || {}
+-var errors = errorObject.bassReferral && errorObject.bassReferral.bassRequest || {}
+
+div.pure-g.borderBottomLight.midPaddingTopBottom
+    div.pure-u-1-2
```
whitfield-mj: This means that on mobile it will be half screen too. Is this what you want? usually you would need .pure-u-1.pure-u-md-1-2 so that it is full width on mob and half on bigger screens

server/routes/bassReferral.js
```diff
@@ -0,0 +1,33 @@
...
+
+    const formConfig = require('./config/bassReferral');
+    const standard = createStandardRoutes({formConfig, licenceService, sectionName: 'bassReferral'});
+
+    router.get('/bassReferral/:formName/:action/:bookingId', async(standard.get));
+    router.post('/bassReferral/:formName/:action/:bookingId', audited, async(standard.post));
+
+    router.get('/bassReferral/:formName/:bookingId', async(standard.get));
+    router.post('/bassReferral/:formName/:bookingId', audited, async(standard.post));
+
```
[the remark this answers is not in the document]
atodd-noms: It's so that you can route back to the check answers page instead of the tasklist after clicking change.

Would need an extra path segment on all the regular routes (ie not change/withdraw/add etc)

server/services/utils/licenceValidation.js
atodd-noms: will come back to it

server/views/review/includes/bassRequest.pug
atodd-noms: done

PR #608, opened 2018-10-25
Ca caselist

whitfield-mj:


test/services/caseListServiceTest.js
```diff
@@ -197,7 +197,21 @@ describe('caseListService', () => {
                     expect(result[0].due).to.eql({text: '6 months', overdue: false});
                 });
```
atodd-noms: Also - Should be one _week_ if 7 days

whitfield-mj: Probably better to stay in days up to 14 days as 10 days seems to be a key one. Probably not granular enough to have 1 week for everything between 7 and 14 days

PR #628, opened 2018-11-08
Combine optout, bass, curfew into one

atodd-noms:


server/routes/address.js
```diff
@@ -80,4 +106,38 @@ function isEmptySubmission(userInput) {
...
+function getCurfewAddressChoice(licence) {
+    if (isYes(licence, ['proposedAddress', 'optOut', 'decision'])) {
+        return 'OptOut';
+    }
+
+    if (isYes(licence, ['proposedAddress', 'addressProposed', 'decision'])) {
+        return 'Address';
+    }
+
+    if (isYes(licence, ['bassReferral', 'bassRequest', 'bassRequested'])) {
```
whitfield-mj: Like the function. We should probably use elsewhere as and when we touch appropriate files

PR #632, opened 2018-11-13
Extract general list handler

atodd-noms:
What do you think of this as a way of generalising handling lists of things eg curfew addresses or BASS requests, in the licence data, so that I can re-use it for handling BASS


server/services/utils/addressHelpers.js
```diff
@@ -1,52 +1,19 @@
-const {getIn, replaceArrayItem, mergeWithRight, removeFromArray} = require('../../utils/functionalHelpers');
+const recordList = require('./recordList');
```
whitfield-mj: I'd recommend doing away with addressHelpers altogether seeing as it is practically empty. Would just put these functions in the service accessing the recordList util

PR #642, opened 2018-11-21
added oauth2 functionality behind feature switch

whitfield-mj:
fixed auth callback use to get. Throw error if required role is not present

Fixed tests

### DIFF
--- a/server/authentication/auth.js
+++ b/server/authentication/auth.js
@@ -1,7 +1,10 @@
 const passport = require('passport');
-const LocalStrategy = require('passport-local').Strategy;
-const {isEmpty} = require('../utils/functionalHelpers');
 const {URLSearchParams} = require('url');
+const strategies = require('./authInit');
+const LocalStrategy = require('passport-local').Strategy;
+const OauthStrategy = require('passport-oauth2').Strategy;
+const config = require('../config');
+const {generateOauthClientToken} = require('./oauth');
 
 function authenticationMiddleware() {
     // eslint-disable-next-line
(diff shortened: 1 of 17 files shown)


server/authentication/auth.js
```diff
@@ -27,22 +30,22 @@ passport.deserializeUser(function(user, done) {
...
-            if (isEmpty(user)) {
-                return done(null, false, {message: 'Incorrect username or password'});
-            }
-
-            return done(null, user);
-        } catch (error) {
-            return done(null, false, {message: 'A system error occured; please try again later'});
-        }
-    });
+function init(signInService, userService, audit, authStrategy) {
```
petergphillips: don't need to pass in `config.authStrategy` here as `auth` already has access to `config`

server/authentication/strategies.js
```diff
@@ -0,0 +1,58 @@
+const fiveMinutesBefore = require('../utils/fiveMinutesBefore');
+const {getIn} = require('../utils/functionalHelpers');
+const allowedRoles = require('./roles');
+const logger = require('../../log');
+
+module.exports = function(signInService, userService, audit) {
```
petergphillips: I wouldn't change it now, but this looks to be over engineered.  Factoring out the strategies into their own file would be good if the feature switch is designed to be long lived and we be using both for a long period of time.  In our case it makes it harder to take the code out when the feature switch is removed.

server/services/userService.js
```diff
@@ -1,6 +1,28 @@
...
+    async function getUserProfile(token, refreshToken, username) {
+        logger.info(`User profile for: ${username}`);
+        const nomisClient = nomisClientBuilder(token);
+
+        const [profile, roles] = await Promise.all([
+            nomisClient.getLoggedInUserInfo(),
+            getAllRoles({token})
+        ]);
+
+        logger.info(`User profile success - token: ${token}`);
```
petergphillips: don't think it is a good idea to log the user's `token` here

server/authentication/strategies.js
whitfield-mj: Discussed and renamed file

server/services/userService.js
whitfield-mj: Agreed

PR #647, opened 2018-11-26
Access RO client credentials in middleware and dont store in cookie

whitfield-mj:


server/authentication/auth.js
```diff
@@ -6,10 +6,13 @@ const OauthStrategy = require('passport-oauth2').Strategy;
...
 const {generateOauthClientToken} = require('./oauth');
 
-function authenticationMiddleware() {
+function authenticationMiddleware(signInService) {
     // eslint-disable-next-line
-    return (req, res, next) => {
+    return async (req, res, next) => {
         if (req.isAuthenticated()) {
+            const {role, username, token} = req.user;
+            res.locals.token = role === 'RO' ? await signInService.getClientCredentialsTokens(username) : token;
```
petergphillips: What magic is going on here?  Doesn't `getClientCredentialsTokens` return an object with the two tokens and expiry time?

whitfield-mj: The magic of making a mistake!

PR #652, opened 2018-11-30
Store previous bass details when rejecting

atodd-noms:


server/routes/bassReferral.js
```diff
@@ -6,40 +6,23 @@ module.exports = ({licenceService}) => (router, audited) => {
 
     const standard = createStandardRoutes({formConfig, licenceService, sectionName: 'bassReferral'});
 
-    router.post('/bassReferral/bassRequest/rejected/:bookingId', audited, asyncMiddleware(replaceBassRequest));
-  
+    router.post('/bassReferral/rejected/:bookingId', audited, asyncMiddleware(async (req, res) => {
+
+        const {bookingId} = req.params;
+        const {enterAlternative} = req.body;
+        const licence = res.locals.licence;
```
whitfield-mj: May aswell be consistent with the two above {licence} = res.locals

server/services/licenceService.js
```diff
@@ -350,6 +351,24 @@ module.exports = function createLicenceService(licenceClient) {
...
+            return licence;
+        }
+
+        const bassRejections = recordList({licence, path: ['bassRejections'], allowEmpty: true});
+        const licenceWithBassRejections = bassRejections.add({record: lastBassReferral});
+
+        const updatedLicence = replacePath(['bassReferral'], {bassRequest: {bassRequested}}, licenceWithBassRejections);
+
+        await licenceClient.updateLicence(bookingId, updatedLicence);
+
```
whitfield-mj: I would just return licenceClient.updateLicence(bookingId, updatedLicence); and then this function doesn't need the async flag

PR #655, opened 2018-12-03
added simplified form validation to eligibility forms

whitfield-mj:


server/utils/functionalHelpers.js
```diff
@@ -145,3 +147,18 @@ function pickKey(predicate, object) {
...
+        R.path(fieldPath)
+    )(fieldConfig);
+}
+
+function getFieldName(fieldConfig) {
+    return R.pipe(
+        R.keys,
+        R.head
+    )(fieldConfig);
+}
```
whitfield-mj: Don't know if these should belong here as they are not generic. Maybe in a formConfigMethods file?

PR #657, opened 2018-12-04
Even further val

whitfield-mj:


server/routes/routeWorkers/standard.js
```diff
@@ -50,24 +46,18 @@ module.exports = ({formConfig, licenceService, sectionName}) => {
             });
 
             if (formConfig[formName].validate) {
-                const errors = licenceService.validateForm(req.body, formConfig[formName]);
+                const errors = licenceService.validateForm(
+                    updatedLicence[sectionName][formName],
```
atodd-noms: updatedLicence[sectionName] could be undefined on the first get

PR #664, opened 2018-12-06
added in place validation to address form

whitfield-mj:


server/services/licenceService.js
```diff
@@ -241,7 +240,7 @@ module.exports = function createLicenceService(licenceClient) {
                 const innerFieldMap = field[fieldName].contains;
                 const innerAnswers = getFormResponse(innerFieldMap, userInput[fieldName]);
 
-                if (!fieldConfig.saveEmpty && allValuesEmpty(innerAnswers)) {
+                if (allValuesEmpty(innerAnswers)) {
```
whitfield-mj: I don't think we need this anymore....

PR #665, opened 2018-12-07
added review validation for eligibility

whitfield-mj:


server/services/utils/formValidation.js
```diff
@@ -11,6 +11,8 @@ const {
     lastItem
 } = require('../../utils/functionalHelpers');
 
+const proposedAddressConfig = require('../../routes/config/proposedAddress');
```
whitfield-mj: Not really happy with accessing config directly, but otherwise the routes/review would need every config file just to pass it through

PR #667, opened 2018-12-10
Lic 790 ro val

whitfield-mj:


server/routes/config/curfew.js
```diff
@@ -56,23 +72,65 @@ module.exports = {
...
+                validationMessage: 'Enter a valid time'
+            }},
+            {sundayFrom: {
+                responseType: 'requiredTime',
+                validationMessage: 'Enter a valid time'
+            }},
+            {sundayUntil: {
+                responseType: 'requiredTime',
+                validationMessage: 'Enter a valid time'
+            }}
```
atodd-noms: Would it be better if there was a default message for each responseType, or is it better to have it all in one place even with the repetition?

PR #676, opened 2018-12-13
Allow no specific bass area option

atodd-noms:


server/routes/routeWorkers/standard.js
```diff
@@ -36,7 +36,7 @@ module.exports = ({formConfig, licenceService, sectionName}) => {
         const validationErrors = updatedLicence => {
             const form = updatedLicence[sectionName][formName];
             // address is in array
-            const formToValidate = form.addresses ? lastItem(form.addresses) : form;
+            const formToValidate = form && form.addresses ? lastItem(form.addresses) : form;
```
whitfield-mj: not sure why form would ever be undefined

PR #688, opened 2018-12-18
display list of refusal reasons if there are multiple

whitfield-mj:


server/utils/licenceStatus.js
```diff
@@ -350,12 +350,13 @@ function getDmApproval(licence) {
...
-    const reason = getIn(licence, ['approval', 'release', 'reason']);
+    const reasons = flatten([getIn(licence, ['approval', 'release', 'reason'])]);
 
     return {
         approved: decision === 'Yes',
         refused: decision === 'No',
-        refusalReason: refusalReasons[reason],
+        refusalReason: reasons.reduce((reasons, reason, index) =>
+            `${reasons}${index > 0 ? ', ' : ''}${index > 0 ? refusalReasons[reason].toLowerCase() : refusalReasons[reason]}`, ''),
         approval: isEmpty(decision) ? taskStates.UNSTARTED : taskStates.DONE
```
atodd-noms: what if you made refusalReasons all lowercase, do a join, and capitalise initial letter?

whitfield-mj: Good call

PR #708, opened 2019-01-03
add message if date is in the past

whitfield-mj:


server/services/utils/formValidation.js
```diff
@@ -134,13 +134,20 @@ function validate({formResponse, pageConfig, formType = 'standard', bespokeCondi
...
+        const errorMessage = getErrorMessage(fieldConfig, error, procedure.getErrorMessage);
 
         const errorObject = error.path.reduceRight((errorObj, key) => ({[key]: errorObj}), errorMessage);
         return mergeWithRight(errors, errorObject);
     }, {});
 }
 
+function getErrorMessage(fieldConfig, error, errorMethod) {
+    if (error.type === 'date.min') {
+        return 'Enter a date that is in the future';
```
petergphillips: I can't really tell, but what happens if you enter today's date?  Is that allowed?

whitfield-mj:
It does reject today, which I assume was specified when it was originally
implemented. Will run past Jim

server/services/utils/formValidation.js
whitfield-mj: It does reject today, which I assume was specified when it was originally
implemented. Will run past Jim

PR #714, opened 2019-01-08
Lic 626 risk management

whitfield-mj:


test/routes/curfewTest.js
```diff
@@ -167,45 +166,33 @@ describe('/hdc/curfew', () => {
...
-                            config: formConfig[route.section],
-                            userInput: route.body,
-                            licenceSection: 'curfew',
-                            formName: route.section
-                        });
-
-                        expect(res.header.location).to.equal(route.nextPath);
-                    });
-            });
+            // it(`renders the correct path '${route.nextPath}' page`, () => {
```
petergphillips: ??? delete delete delete

server/views/proposedAddress/rejected.pug
```diff
@@ -33,8 +33,9 @@ block content
                 p.smallMarginBottom Reason for rejection:
                 if address.consent === 'No'
                     p.bold The homeowner/landlord does not give informed consent
+                // TODO check this
```
petergphillips: have you checked this :point_up: .  If it should still be there then put your name against it in the TODO

PR #735, opened 2019-01-22
Lic 768 versioning2

whitfield-mj:


server/routes/viewModels/taskListModels.js
```diff
@@ -124,8 +124,8 @@ module.exports = (
         addressRejectedInRiskTask: addressUnsuitable,
         addressOrBassOfferedOrUnsuitable: curfewAddressApproved || bassOfferMade || addressUnsuitable,
         licenceUnstarted: stage === 'UNSTARTED',
-        licenceVersionExists: approvedVersion,
-        licenceVaried: !approvedVersion || version > approvedVersion
+        licenceVersionExists: !isEmpty(approvedVersionDetails),
+        licenceVaried: isEmpty(approvedVersionDetails) || version > approvedVersion
```
petergphillips: Not following this, a licence is varied if the version is greater than the approved?  I thought the vary version was a separate field?

whitfield-mj: Yes but this version / approved version is the result of sticking them together e.g. 1.2, 1.1
See licenceService line 55 / 56

petergphillips: but:
```            const version = `${rawLicence.version}.${rawLicence.vary_version}`;
``` looks like a string to me?  Won't that mean that `1.9 > `1.10`?

Comment (whitfield-mj) [reply server/routes/viewModels/taskListModels.js]: Correct, need to compare the numbers separately as we are using semantic versioning rather than normal numbers

File: server/routes/viewModels/taskListModels.js [hunk 2]
```diff
@@ -159,3 +159,8 @@ function getBassDetails({bassReferralNeeded, bassAccepted, bassWithdrawn}, {bass
         bassOfferMade: bassReferralNeeded && bassOffer === 'DONE' && !bassWithdrawn && !bassExcluded
     };
 }
+
+function versionAheadOfApproved(versionDetails, approvedVersionDetails) {
+    return versionDetails.version > approvedVersionDetails.version ||
```
petergphillips: looks fine _provided_ that once they have started varying the licence version will never change i.e. we won't get versions of 2.4 and 4.2 to compare...

whitfield-mj: Agreed

PR #740, opened 2019-01-24
capitalise all usernames. Log and signout if error getting admin token

whitfield-mj:


server/authentication/authInit.js
```diff
@@ -5,12 +5,13 @@ const logger = require('../../log');
 module.exports = function(signInService, userService, audit) {
 
     async function localInit(username, password, done) {
+        const capUsername = username.toUpperCase();
```
petergphillips: confused as to why you are changing this function.  Shouldn't it just be deleted as all logins should go through auth?

server/authentication/auth.js
```diff
@@ -15,9 +16,14 @@ function authenticationMiddleware(signInService) {
...
-            const systemToken = await signInService.getClientCredentialsTokens(username);
-            res.locals.token = systemToken.token;
-            return next();
+            try {
+                const systemToken = await signInService.getClientCredentialsTokens(username);
+                res.locals.token = systemToken.token;
+                return next();
+            } catch (error) {
+                logger.error('Unable to get client credentials token for RO', error.stack);
+                return res.redirect('/logout');
```
petergphillips: is the right place to redirect them?  Won't they logout, login again and then be logged out automatically?

server/authentication/authInit.js
whitfield-mj: Yes everything that a user sees. It was just for instances when running locally with local signin form

server/authentication/auth.js
whitfield-mj: Yes. Not an ideal solution. Shouldn't really occur in the wild. Maybe this full PR is unnecessary, will close

PR #753, opened 2019-02-05
Dm task list

whitfield-mj:


server/routes/viewModels/taskListModels.js
```diff
@@ -48,7 +49,12 @@ const tasksConfig = {
     caTasksFinalChecks: [
         {task: 'curfewAddressTask', filters: ['!bassReferralNeeded', '!caToRo']},
         {task: 'proposedAddressTask', filters: ['caToRo']},
-        {task: 'bassOfferTask', filters: ['bassReferralNeeded']},
+        {
+            title: 'BASS address',
```
petergphillips: would be clearer if the fields were always in the same order, the last PR had `filters` first, then `title`

PR #755, opened 2019-02-05
Dm task list risk

whitfield-mj:


server/routes/viewModels/taskLists/tasks/bassOffer.js
```diff
@@ -22,9 +22,9 @@ module.exports = {
             }
             // TODO warning box
```
petergphillips: have you done this TODO now?

PR #758, opened 2019-02-06
Add push to nomis capability. Send approved/rejected on decision from DM

whitfield-mj:


server/routes/approval.js
```diff
@@ -20,7 +18,7 @@ module.exports = ({licenceService, prisonerService}) => (router, audited) => {
 
             const {nextPath, pageDataMap} = formConfig[formName];
             const dataPath = pageDataMap || ['licence', 'approval', 'release'];
-            const data = getIn(res.locals.licence, dataPath) || {};
+            const data = firstItem(req.flash('userInput')) || getIn(res.locals.licence, dataPath) || {};
```
petergphillips: what's going on here?

server/services/nomisPushService.js
```diff
@@ -0,0 +1,29 @@
+const logger = require('../../log.js');
+
+module.exports = nomisClientBuilder => {
+
+    async function pushStatus(bookingId, approvalDecision, systemToken) {
+        const nomisClient = nomisClientBuilder(systemToken);
+        const status = (() => {
```
petergphillips: don't really understand this.  You creating a function to work out the status and then calling the function immediately?  Why not just have a case statement, why the need to wrap in a dynamic function?

test/supertestSetup.js
```diff
@@ -19,11 +19,11 @@ const auditStub = {
     record: sinon.stub()
 };
 
-const signInServiceStub = {
+const createSignInServiceStub = () => ({
```
petergphillips: why the change?

server/routes/approval.js
whitfield-mj: If the page is redirected to after a validation error then the data to fill the form fields with will be in req.flash. If it is accessed having already been inputted it will be in the licence. Otherwise empty

server/services/nomisPushService.js
whitfield-mj: Just another way of doing it. Not uncommon to have a self invoking function in JS. Happy to use Switch if preferred

test/supertestSetup.js
whitfield-mj: Most of the others now use the create method. Having one that is used across tests caused the test suite to hang when I called a reset

server/services/nomisPushService.js
petergphillips: I would prefer:
```javascript
        const bob = {
            Yes: 'Approved',
            No: 'Rejected'
        }
        const status = bob[approvalDecision]
```

test/supertestSetup.js
petergphillips: :+1:

PR #759, opened 2019-02-07
removed licenceConditionTask view file. moved logic to task list models

whitfield-mj:


server/routes/viewModels/taskLists/tasks/additionalConditions.js
```diff
@@ -0,0 +1,22 @@
...
+    getLabel: ({decisions, tasks}) => {
+        const {standardOnly, bespoke, additional} = decisions;
+        const {licenceConditions} = tasks;
+
+        if (licenceConditions === 'DONE') {
+            if (standardOnly) {
+                return 'Standard conditions only';
+            }
+            const totalConditions = bespoke + additional;
+            return totalConditions > 1 ? `${totalConditions} conditions added` : `${totalConditions} condition added`;
```
petergphillips: wasn't clear what was going on here, I think:
```javascript
return `${totalConditions} condition${totalConditions > 1 ? 's' : ''} added`
```
would be clearer

PR #766, opened 2019-02-08
dont show caseload in header if user doesnt have one

whitfield-mj:


server/views/global-header.pug
```diff
@@ -14,7 +14,7 @@ header(role="banner", id="global-header", class="with-proposition")
                     div#header-user.u-pad10.sm-pad15
                         div.u-padRight10.sm-padRight30.padLeft15.bold.borderRight
                             a(href="/user" title="User page" aria-label="Visit user admin page") #{user.firstName} #{user.lastName}
-                        if user.activeCaseLoad
+                        if user.activeCaseLoad && !user.activeCaseLoad.description.includes('---')
                             div.u-padRightLeft10.sm-padLeftRight30.bold.borderRight
```
whitfield-mj: Not a very nice way of doing it but it seems we get 
` activeCaseLoad: 
   { caseLoadId: '___',
     description: '-------',
     type: 'DUMMY',
     caseloadFunction: 'GENERAL' } }`

If there is no caseload

PR #790, opened 2019-02-19
LIC-881: Switch to oauth2 for feature tests

petergphillips:


licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/pdf/PdfSpec.groovy
```diff
@@ -37,6 +37,7 @@ class PdfSpec extends GebReportingSpec {
 
     then: 'There is no option to create PDF'
     !taskListAction('Create licence').isDisplayed()
+    actions.logOut()
```
petergphillips: needed as the test tries to login as two people without logging out.  This doesn't work when auth remembers who you are after login

server/authentication/signInService.js
```diff
@@ -60,14 +38,17 @@ function signInService() {
...
+  const token = oauthResult.body.access_token
+  const refreshToken = oauthResult.body.refresh_token
+  const expiresIn = oauthResult.body.expires_in
 
-  return parseOauthTokens(oauthResult)
+  return { token, refreshToken, expiresIn }
 }
 
-function getOauthToken(oauthClientToken, requestSpec) {
+const getOauthUrl = () => config.nomis.authUrl
```
whitfield-mj: Seems unnecessary to have this function

PR #791, opened 2019-02-20
LIC-881: Fix staging feature tests and tidy auth code

petergphillips:


licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/pages/SigninPage.groovy
```diff
@@ -14,10 +14,10 @@ class SigninPage extends Page {
   static content = {
     signInAs { user ->
       $('form').username = user
-      $('form').password = 'licences123456'
+      $('form').password = 'licences123457'
```
petergphillips: note stage password change due to expired password

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/pages/SigninPage.groovy
```diff
@@ -7,17 +7,19 @@ class SigninPage extends Page {
 
   static url = '/login'
 
+  static password = System.getenv('USER_PASS') ?: 'licences123456'
```
petergphillips: had to add to env variables as different password for local and stage

PR #794, opened 2019-02-25
removed expensive romanise calculation and added lookup table

whitfield-mj:


server/utils/romanise.js
```diff
@@ -1,31 +1,58 @@
...
-  return convert(number)
-}
-
-function convert(number) {
-  return (
-    firstItem(
-      numerals.filter(([arabic]) => arabic <= number).map(([arabic, roman]) => roman + convert(number - arabic))
-    ) || ''
-  )
+  return conversions[number]
```
petergphillips: Is it worth adding a test that uses romanise to convert back again, so that you know the lookup is correct?  Definitely worth testing what happens with numbers like 12 and 18 :grinning:

server/services/utils/pdfFormatter.js
```diff
@@ -1,7 +1,7 @@
 const moment = require('moment')
 const { getIn, isEmpty, mergeWithRight } = require('../../utils/functionalHelpers')
 const pdfData = require('../config/pdfData')
-const { romanise } = require('../../utils/romanise')
+const romanise = require('romannumerals')
```
petergphillips: odd, I would have expected this to have failed linting since the `require` is after local `require` statements

PR #804, opened 2019-02-28
LIC-878: Use currently active from caseloads endpoint instead

petergphillips:


server/services/userService.js
```diff
@@ -10,14 +10,16 @@ module.exports = nomisClientBuilder => {
 
     logger.info(`User profile success - username: ${username}`)
 
-    const activeCaseLoads = profile.activeCaseLoadId ? await nomisClient.getUserCaseLoads() : []
-    const activeCaseLoad = activeCaseLoads.find(caseLoad => caseLoad.caseLoadId === profile.activeCaseLoadId)
+    const activeCaseLoads = await nomisClient.getUserCaseLoads()
+    const activeCaseLoad = activeCaseLoads.find(cl => cl.currentlyActive)
+    const activeCaseLoadId = activeCaseLoad ? activeCaseLoad.caseLoadId : undefined
```
whitfield-mj: Could use `const activeCaseLoadId =  getIn(activeCaseload, ['caseLoadId'])` Not sure if it is any better though

PR #809, opened 2019-03-01
removed git tagging from deploy_stage step

whitfield-mj:


.circleci/config.yml
```diff
@@ -187,10 +187,6 @@ jobs:
             VERSION=$(egrep buildNumber build-info.json | awk -F"\"" '{print $4}')
             npm run plant-beanstalk ${VERSION}
             GIT_REF=$(egrep gitRef build-info.json  | awk -F"\"" '{print $8}')
```
Mjwillis: don't need GIT_REF

whitfield-mj: removed

PR #810, opened 2019-03-01
get releasing establishment for prisoners who have been released

whitfield-mj:


server/services/prisonerService.js
```diff
@@ -70,17 +70,11 @@ function createPrisonerService(nomisClientBuilder) {
...
+    logger.info(`getPrisonerImage: ${imageId}`)
 
-      const nomisClient = nomisClientBuilder(token)
-      const image = await nomisClient.getImageData(imageId)
-      return image
-    } catch (error) {
-      logger.error('Error getting prisoner image')
-      return null
-    }
+    const nomisClient = nomisClientBuilder(token)
```
petergphillips: out of interest, why the code change here?

server/services/prisonerService.js
```diff
@@ -70,17 +70,11 @@ function createPrisonerService(nomisClientBuilder) {
     return [merge(coms[0], { deliusId: id.identifierValue })]
   }
 
-  async function getPrisonerImage(imageId, token) {
-    try {
-      logger.info(`getPrisonerImage: ${imageId}`)
+  function getPrisonerImage(imageId, token) {
```
petergphillips: do you prefer `function` over `const getPrisonerImage` - just wondered why not changed to arrow syntax?

server/services/prisonerService.js
whitfield-mj: The linter highlighted defining image just to return it on the next line and while I was there I noticed that it’s pointless to do so. An async function returns a promise so all we’re doing is resolving a promise just to wrap it in one again. May as well pass the original promise straight through

whitfield-mj:
In terms of function vs const. never defined it and it uses both throughout. Would be a big job to change all. I also prefer the hoisting characteristics of function at times.

PR #829, opened 2019-03-15
added new caselist status labels. Refactored filtering for tabs

whitfield-mj:


server/services/utils/caseListFormatter.js
```diff
@@ -34,11 +34,12 @@ function getStatus(prisoner, licences, role) {
   })
 
   if (!licenceForPrisoner) {
-    return { stage: licenceStages.UNSTARTED, status: 'Not started' }
+    return { stage: licenceStages.UNSTARTED, status: 'Not started', activeCase: true }
   }
 
   const licenceStatus = getLicenceStatus(licenceForPrisoner)
-  return { stage: licenceForPrisoner.stage, status: getStatusLabel(licenceStatus, role) }
+  const { statusLabel, activeCase } = getStatusLabel(licenceStatus, role)
```
whitfield-mj: Maybe something to look at another time. Now getStatusLabel returns whether it is active too would it be better to incorporate into the getLicenceStatus function called immediately above? The active-ness of the licence has always been strongly linked to the label (it is how it was defined originally) but maybe seems out of place?

atodd-noms: getLicenceStatus is a bit of a monster. Need to find a way to simplify that first, then combine. Will put it on the list.

PR #831, opened 2019-03-19
Support multiple DM emails/names in notification config

atodd-noms:
Also move notification data code out of send route


server/services/config/notificationMailboxes.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  LT1: {
+    ca: ['hdc_test@digital.justice.gov.uk', 'hdc_test+2@digital.justice.gov.uk'],
+    dm: [
+      { email: 'hdc_test@digital.justice.gov.uk', name: 'DM Test1' },
+      { email: 'hdc_test+2@digital.justice.gov.uk', name: 'DM Test2' },
+    ],
+  },
+  DTI: {
```
petergphillips: is this temporary?  Doesn't look the correct place for environmental configuration...

atodd-noms: yeah, temporary until a story for a full DB solution for long list of prisons. These are just 3 prisons active in initial trial

PR #837, opened 2019-03-21
Restore no-console lint rule

atodd-noms:
Put back previously agreed team rules

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,8 @@
     "prettier/prettier": ["error"],
     "no-use-before-define": "off",
     "prefer-destructuring": "warn",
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -4,7 +4,8 @@
     "prettier/prettier": ["error"],
     "no-use-before-define": "off",
     "prefer-destructuring": "warn",
-    "consistent-return": "off"
+    "consistent-return": "off",
+    "no-console": ["error"]
```
petergphillips: it would be good to understand why you need to expicitly put this in.  We have a no-console rule in notm, but haven't had to put it in .eslintrc.  What other rules are missing?

PR #839, opened 2019-03-22
Pass domain through to new versions of notification templates

atodd-noms:
add domain to the notification data, added corresponding domain placeholder in the URLs in the templates instead of hardcoded links

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -9,7 +9,7 @@ const userClient = require('./data/userClient')
 const nomisClientBuilder = require('./data/nomisClientBuilder')
 const pdfFormatter = require('./services/utils/pdfFormatter')
 
-const notifyClient = new NotifyClient(config.notifications.notifyKey)
+const notifyClient = new NotifyClient(config.notificationConfig.notifyKey)
 const createSignInService = require('./authentication/signInService')
 const createLicenceService = require('./services/licenceService')
 const { createPrisonerService } = require('./services/prisonerService')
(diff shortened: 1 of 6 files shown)


server/index.js
```diff
@@ -9,7 +9,7 @@ const userClient = require('./data/userClient')
 const nomisClientBuilder = require('./data/nomisClientBuilder')
 const pdfFormatter = require('./services/utils/pdfFormatter')
 
-const notifyClient = new NotifyClient(config.notifications.notifyKey)
+const notifyClient = new NotifyClient(config.notificationConfig.notifyKey)
```
petergphillips: not sure if `config.notificationConfig` is any better than just `config.notification`

server/services/notificationService.js
```diff
@@ -1,6 +1,6 @@
 const templates = require('./config/notificationTemplates')
 const notificationMailboxes = require('./config/notificationMailboxes')
-const { notifyKey } = require('../config').notifications
+const { notificationConfig, domain } = require('../config')
```
petergphillips: would be better to destructure this instead i.e. something like
```const { notifications: { notifyKey }, domain } = require('../config')
```
and then you can just refer to the `notifyKey` as before

server/index.js
atodd-noms: was to avoid name conflicts but your other suggestion fixes the problem ta

server/services/notificationService.js
atodd-noms: nice, ta

PR #841, opened 2019-03-25
Add NOMS number to notifications

atodd-noms:
Also remove unnecessary API call because prisoner personal details already in request scope


atodd-noms:
Yes, that's it. I didn't know it happened, but I noticed that in the middleware on each request we already retrieve the prisoner details and put it in res.locals, because it's used on the page header on most pages. I was calling it a second time without needing to

PR #843, opened 2019-03-26
DB storage of notification mailboxes & admin UI to manage them

atodd-noms:
NB this is only reachable by admin users and sits alongside RO user management

PS what was in server/routes/admin/admin.js is now in server/routes/admin/users.js alongside server/routes/admin/mailboxes.js


migrations/20190325115901_create-notifications-config.js
```diff
@@ -0,0 +1,13 @@
...
+  Promise.all([
+    knex.schema.createTable('notifications_config', table => {
+      table.increments('id').primary('pk_notifications')
+      table.string('email').notNullable()
+      table.string('establishment').notNullable()
+      table.string('role').notNullable()
+      table.string('name').nullable()
+      table.unique(['email', 'establishment', 'role'])
+    }),
+  ])
```
timharrison-moj: How is this applied to the database? And if you needed to change the structure of a table, how might you do this without needing to recreate the table?

server/data/configClient.js
```diff
@@ -0,0 +1,66 @@
+const db = require('./dataAccess/db')
+
+module.exports = {
+  async getAllMailboxes() {
+    const query = {
+      text: 'select * from notifications_config order by establishment ASC',
+    }
```
timharrison-moj: Would it be better to select specific columns?  Any change in table structure would also change the object produced from these queries. Its fine if that is the intention though.

server/data/configClient.js
```diff
@@ -0,0 +1,66 @@
...
+  async addMailbox({ email, establishment, role, name }) {
+    const query = {
+      text: `insert into notifications_config
+                (email, establishment, role, name)
+                values($1, $2, $3, $4)`,
+      values: [email, establishment, role, name],
+    }
+
+    return db.query(query)
+  },
```
timharrison-moj: What happens if there are data access exceptions?  E.g. database connection errors, missing table errors? Are those cleanly caught and handled with a message to the user, or silently handled?

test/data/configClientTest.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  const mailbox2 = {
+    id: 'id2',
+    establishment: 'establishment2',
+    email: 'email2',
+    name: 'name2',
+    role: 'role2',
+  }
+
```
timharrison-moj: Does this limit the number of emails addresses to 2? I know some staff could potentially have multiple email addresses in NOMIS. If only 2 are needed, its fine.

test/routes/mailboxesAdminTest.js
```diff
@@ -0,0 +1,246 @@
...
+        .post('/admin/mailboxes/add/')
+        .send({ establishment: '', email: 'email1', role: 'role1', name: 'name1' })
+        .expect(302)
+        .expect('Location', '/admin/mailboxes/add')
+        .expect(() => {
+          expect(configClient.addMailbox).not.to.be.calledOnce()
+        })
+    })
+
+    it('redirects back to page and does not call config client when role not CA or DM', () => {
```
timharrison-moj: These tests are using a limited data set to try the features out. Might benefit from a wider range of data?

test/routes/userAdminTest.js
```diff
@@ -49,16 +49,6 @@ describe('/admin', () => {
...
-  describe('GET /admin', () => {
-    it('redirects to ro user list', () => {
-      const app = createApp({ userAdminServiceStub: userAdminService }, 'batchUser')
-      return request(app)
-        .get('/admin/')
-        .expect(302)
-        .expect('Location', '/admin/roUsers')
-    })
-  })
-
```
timharrison-moj: Same comment as above - just limited data used in the test cases. If one of queries was more 'open' than it ought to be there is a chance that the test still passes but in real environments the additional data could cause it to fail.

migrations/20190325115901_create-notifications-config.js
romidane: @timharrison-moj this is a migration file.. and if you needed to change a the table structure you'd have to create a new migration file with an alter command... something like `knex.schema.alterTable('xyz')`
The way migrations work is that it creates a migrations table which stores the last migration that was run.
If it's been run already run it won't run it again... so you don't actually have to recreate the table..

assets/javascripts/listSort.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable */
+
+$(document).ready(init)
```
romidane: any point having the `init` function as part of the global scope? probably worth just sticking it in the ready as a callback?

migrations/20190325115901_create-notifications-config.js
atodd-noms: We use knex.js for migrations. Future migrations would be non-destructive.
Small data set now but added index for future

server/data/configClient.js
atodd-noms: agreed

server/data/configClient.js
atodd-noms: They will be handled by the route and for certain cases shown in-place on the form, other general errors like timeouts will route to the catch-all error handler (Something went wrong, try again)

test/data/configClientTest.js
atodd-noms: No, just a mock response. Not even used really because we're only testing the inputs, this response is just a dummy. However, looking closer revealed it wouldn't handle an empty result set so I've added a test for that.

test/routes/mailboxesAdminTest.js
atodd-noms: Added some cases to make more explicit that any input is allowed regardless of format as long as mandatory fields present

test/routes/userAdminTest.js
atodd-noms: Added more cases to make it more explicit that any input is accepted

assets/javascripts/listSort.js
atodd-noms: Nope, no point. Have brought it inline

PR #846, opened 2019-03-29
Ensure notify errors don't interrupt sending

atodd-noms:
Also clean up send route by moving to notification service.

Notifications are non-critical so if an error occurs it only needs to be logged but should not cause the handover to fail. Therefore have re-ordered code so that errors during notify will not affect send.

Also if the handover fails then notifications should not be sent, so now the order ensures that too.

### DIFF
--- a/server/routes/send.js
+++ b/server/routes/send.js
@@ -1,4 +1,5 @@
 const { asyncMiddleware } = require('../utils/middleware')
+const logger = require('../../log.js')
 
 module.exports = ({ licenceService, prisonerService, notificationService, audit }) => router => {
   router.get('/:destination/:bookingId', async (req, res) => {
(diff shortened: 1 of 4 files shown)


server/routes/send.js
```diff
@@ -25,31 +26,30 @@ module.exports = ({ licenceService, prisonerService, notificationService, audit
...
-      ])
+      await licenceService.markForHandover(bookingId, transition.type)
 
       if (transition.type === 'dmToCaReturn') {
         await licenceService.removeDecision(bookingId, res.locals.licence)
       }
 
       auditEvent(req.user.username, bookingId, transition.type, submissionTarget)
 
+      try {
```
romidane: I'd say instead of having the try catch here perhaps it's better to have it inside the `sendNotifications` method have that deal with the error silently.. you could even test that if you wanted to?

atodd-noms: Good point. Will fix

PR #847, opened 2019-04-01
Services to find due/overdue cases and send notifications

atodd-noms:
Sorry there's quite a bit of code in this. Two main bits - find all the cases where the dates mean they are due or due soon or overdue. There's some working day calculations required to enable this. Then send reminder notifications for each of those cases. It will be triggered by an overnight batch job in the early hours, which I've split into a subsequent PR.

I've purposely made the sending part sequential because with similar batch jobs in the batchloader app we found that it quickly caused timeouts from the nomis API. The volumes here are hard to predict but although it's quite likely there could be enough to cause timeouts if we bombard the nomis api, there won't be so many that it would take too long for an over night batch job.

### DIFF
--- a/server/services/deadlineService.js
+++ b/server/services/deadlineService.js
@@ -16,23 +16,13 @@ module.exports = function createDeadlineService(licenceClient) {
     }
     rejectIfUnknown(role)
     const { from, upto } = dueDateMethods[role].due(days)
-    try {
-      return licenceClient.getLicencesInStageBetweenDates(roleStages[role], from, upto)
-    } catch (error) {
-      logger.error('Error getting due licences', error.stack)
-      throw error
-    }
+    return licenceClient.getLicencesInStageBetweenDates(roleStages[role], from, upto)
   }
 
   async function getOverdue(role) {
     rejectIfUnknown(role)
     const upto = dueDateMethods[role].overdue()
-    try {
-      return licenceClient.getLicencesInStageBeforeDate(roleStages[role], upto)
-    } catch (error) {
-      logger.error('Error getting overdue licences', error.stack)
-      throw error
-    }
+    return licenceClient.getLicencesInStageBeforeDate(roleStages[role], upto)
   }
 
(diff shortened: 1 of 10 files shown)


server/services/deadlineService.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+module.exports = function createDeadlineService(licenceClient) {
+  async function getDueInDays(role, days) {
+    if (!Number.isInteger(days) || days < 0) {
+      throw new Error('Days must be a whole number')
+    }
+    rejectIfUnknown(role)
+    const { from, upto } = dueDateMethods[role].due(days)
+    try {
```
romidane: Curious.. Is there any need to wrap this in a try catch simply so you can log an error..
The method `getDueInDays ` and  erroring method `getLicencesInStageBetweenDates` is well named so it would be easy to find and debug from the logs?

atodd-noms: No, there isn't really. I was just being consistent with most of the other service code which does the same thing, also without any real need to. I'll take it out here. Probably nees to come out in lots of other old code too, eventually.

test/data/licenceClientTest.js
```diff
@@ -309,4 +309,64 @@ describe('licenceClient', () => {
...
     })
   })
+
+  describe('getLicencesInStageBetweenDates', () => {
+    it('should call query', () => {
+      licencesProxy().getLicencesInStageBetweenDates('stage', 'from', 'upto')
+      expect(queryStub).to.have.callCount(1)
+    })
+
+    it('should pass in the correct sql', () => {
```
romidane: Curious.. This test seems like it's testing hard coded values from the method and from the looks of it is only giving you assurance that the sql query didn't change... is this the intention?

atodd-noms: No, that's pretty pointless. Following the pattern matt established in the rest of that test. Will remove, there's nothing dynamic in that query

PR #853, opened 2019-04-05
Lic 834 checks passed alternate

atodd-noms:


server/routes/routeWorkers/standard.js
```diff
@@ -95,9 +73,54 @@ module.exports = ({ formConfig, licenceService, sectionName, nomisPushService, c
     res.redirect(`${nextPath}${bookingId}`)
   }
 
+  function getTarget(formName) {
+    const saveSection = formConfig[formName].saveSection || []
+    const targetSection = saveSection[0] || sectionName
```
romidane: where you getting sectionName from?

atodd-noms: At creation. Line 4:

`module.exports = ({ formConfig, licenceService, sectionName, nomisPushService, config }) => {
  function get(req, res) {`

PR #857, opened 2019-04-10
Required RO user management

atodd-noms:
Adds a service to lookup incomplete users out of the audit data so that internal admins (Jim) can onboard new users.

Lots of this is just pug view files for the admin screens. Lots of it is tests.

Only used by internal admins, not user facing.

PS finding "incomplete" users means looking in the audit for users that have been sent cases (action: SEND, transitionType:caToRo), comparing those to the staff_ids table which holds all onboarded users, and if the user is missing, take the first time they appeared in the audit. Also have to call nomis to lookup the offender nomis ID matching that case.

After that, just show the list of incomplete users in a table.

### DIFF
--- a/server/data/userClient.js
+++ b/server/data/userClient.js
@@ -1,3 +1,4 @@
+const moment = require('moment')
 const db = require('./dataAccess/db')
 
 module.exports = {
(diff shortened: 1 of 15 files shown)


server/routes/admin/users.js
```diff
@@ -127,5 +127,25 @@ module.exports = ({ userAdminService }) => (router, audited) => {
...
+      const input = req.body.incompleteUser
+      const userInput = JSON.parse(input)
+      req.flash('userInput', userInput)
+
+      return res.redirect('/admin/roUsers/add')
+    } catch (error) {
+      return res.redirect('/admin/roUsers/incomplete')
+    }
+  })
+
```
atodd-noms: This is redirecting to the existing "add user" page, passing the names and ID we already have so that you don't have to type it in again.

server/services/userAdminService.js
```diff
@@ -108,11 +111,57 @@ module.exports = function createUserService(nomisClientBuilder, userClient) {
...
+
+  async function getOffenderNomis(token, bookingIds) {
+    try {
+      const nomisClient = nomisClientBuilder(token)
+      const offenders = await nomisClient.getOffenderSentencesByBookingId(bookingIds, false)
+
+      if (isEmpty(offenders)) {
+        return null
+      }
+
```
atodd-noms: make a map of bookingId to offender number so that we can add offender number to the "incomplete user" object

server/data/userClient.js
```diff
@@ -26,6 +27,40 @@ module.exports = {
...
+             from (
+                    select *,
+                           row_number() over (
+                             partition by (a.details -> 'submissionTarget' -> 'com' -> 'deliusId')
+                             order by a.timestamp ASC
+                             ) as rownbr
+                    from audit a
+                           left join staff_ids s
+                                     on s.staff_id = (a.details-> 'submissionTarget' -> 'com' ->> 'deliusId')
+                    where a.action = 'SEND'
```
timharrison-moj: What are the -> and ->> operators here? Not come across these before... I could google..

atodd-noms: it's postgres JSON. Querying inside a json object. Quite ugly

PR #858, opened 2019-04-11
Case insensitive user search, and handle empty results

atodd-noms:


server/data/userClient.js
```diff
@@ -23,29 +23,41 @@ module.exports = {
...
     }
 
     const { rows } = await db.query(query)
 
-    return convertPropertyNames(rows[0])
+    if (rows && rows[0]) {
+      return convertPropertyNames(rows[0])
+    }
+
+    return undefined
```
romidane: I have a small problem with returning `undefined` from functions.. I know you've tested it but it makes it hard to know whether there was a `noop` or a misspelled variable ect... what are you thoughts on throwing a custom error?

atodd-noms: What about null? It's not an error, just a 0 matches. Empty list where the result would be a list, but what's preferred in javascript land for empty single value?

romidane: Null works.. it's more intentional.. also shows the absence of a value

PR #862, opened 2019-04-16
Use current RO assignements as basis for incomplete user lookup

atodd-noms:
Need to find unmapped RO users by checking current RO assignments, not the assignments when first appeared.

1. Find cases that may require a new RO (ie new cases, ie ELIGIBILITY or PROCESSING_RO)
2. For each case, lookup the assigned RO from nomis
3. Find the set of unique ROs (by delius ID)
4. Check the staff user table for that delius ID and return details if not mapped or mapped but not onboarded
5. Find the offender nomis ID related to each incomplete user


server/data/userClient.js
```diff
@@ -16,35 +15,15 @@ module.exports = {
...
-                  ) s
-             where rownbr = 1`,
+      text: `select booking_id from licences where stage in ('ELIGIBILITY', 'PROCESSING_RO')`,
     }
 
     const { rows } = await db.query(query)
 
     if (rows) {
-      return rows.map(convertIncompleteUserPropertyNames)
+      return rows.map(r => r.booking_id).filter(r => r)
```
romidane: I don't really understand the intention for filter here.. Can you explain?

atodd-noms: Just to remove empty rows but now I realise that shouldn't happen from this query anyway. Will remove

server/services/userAdminService.js
```diff
@@ -111,49 +111,92 @@ module.exports = function createUserService(nomisClientBuilder, userClient) {
...
+        } catch (error) {
+          logger.warn(`Error in getCom for incomplete users, booking id ${bookingId}`, error.stack)
+        }
+      })
+    )
 
-      if (isEmpty(offenders)) {
-        return null
-      }
+    return required.filter(e => e)
```
romidane: not sure what you are filtering out here is `e` a boolean?

server/services/userAdminService.js
```diff
@@ -111,49 +111,92 @@ module.exports = function createUserService(nomisClientBuilder, userClient) {
...
+        } catch (error) {
+          logger.warn(
+            `Error in getRoUserByDeliusId for incomplete users, assigned Id ${assignment.assignedId}`,
+            error.stack
+          )
+        }
+      })
+    )
+
+    return incomplete.filter(e => e)
```
romidane: Not sure what you are filtering here

server/services/userAdminService.js
atodd-noms: Can be anything - just removes where null/undefined

server/services/userAdminService.js
atodd-noms: removing any undefined from the result of map. For each item in the list we get an object if the user is incomplete, otherwise nothing, so the output of the map could have undefineds mixed in

server/data/userClient.js
romidane: I see perhaps you can do it this way to show your intention by writing it as follows:
```
(r => r.booking_id).filter(Boolean)
```

This way all falsy values are cast into booleans?

server/services/userAdminService.js
romidane:
```
return incomplete.filter(Boolean)
```

server/services/userAdminService.js
romidane:
```
return required.filter(Boolean)
```

server/services/userAdminService.js
atodd-noms: OK, cool, I didn't know you could do that

PR #865, opened 2019-04-18
Merge pull request #864 from ministryofjustice/LIC-953-RO-lookup-error

atodd-noms:
Allow passing more meaningful error messages in caselist lookup

### DIFF
--- a/server/services/prisonerService.js
+++ b/server/services/prisonerService.js
@@ -1,8 +1,10 @@
 const logger = require('../../log.js')
 const { formatObjectForView } = require('./utils/formatForView')
-const { merge, getIn } = require('../utils/functionalHelpers')
+const { getIn } = require('../utils/functionalHelpers')
 
 module.exports = { createPrisonerService }
+
+class RoRelationshipError extends Error {}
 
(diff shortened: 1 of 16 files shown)


server/services/prisonerService.js
```diff
@@ -59,15 +59,95 @@ function createPrisonerService(nomisClientBuilder) {
     }
   }
 
-  async function getComDetails(nomisClient, coms) {
-    if (!coms[0]) {
-      return null
+  async function getResponsibleOfficer(bookingId, token) {
+    logger.info(`getResponsibleOfficer: ${bookingId}`)
+    const com = await findResponsibleOfficer(bookingId, token)
```
romidane: You throw an error for unsuspected errors in the `findResponsibleOfficer` and you are not dealing with that error here.. is the goal to let the error bubble up to the next caller?

atodd-noms: Yeah, all the ones we can handle (for data issues we know about) have been turned into a message, but anything unexpected like network failures will bubble up to the global handler and the standard error page.

PR #866, opened 2019-04-24
Push checks passed on failure

atodd-noms:
Checks passed is also used for false, not just true
Add route config to trigger checks failed push where ever needed
Change input to objects for consistency and to avoid passing unlabelled boolean
Make username consistent with prevailing usage

### DIFF
--- a/server/data/nomisClientBuilder.js
+++ b/server/data/nomisClientBuilder.js
@@ -3,7 +3,7 @@ const superagent = require('superagent')
 const logger = require('../../log')
 const config = require('../config')
 const { merge, pipe, getIn } = require('../utils/functionalHelpers')
-const { NoTokenError } = require('../utils/errors')
+const { unauthorisedError } = require('../utils/errors')
 
 const timeoutSpec = {
   response: config.nomis.timeout.response,
@@ -142,6 +142,10 @@ module.exports = token => {
     },
 
     async putChecksPassed({ bookingId, passed }) {
+      if (typeof passed !== 'boolean') {
+        throw new Error(`Missing required input parameter 'passed'`)
+      }
+
       const path = `${apiUrl}/offender-sentences/booking/${bookingId}/home-detention-curfews/latest/checks-passed`
       const body = { passed, date: moment().format('YYYY-MM-DD') }
 
@@ -159,7 +163,7 @@ module.exports = token => {
 function nomisGetBuilder(token) {
   return async ({ path, query = '', headers = {}, responseType = '' } = {}) => {
     if (!token) {
-      throw new NoTokenError()
+      throw unauthorisedError()
     }
 
     try {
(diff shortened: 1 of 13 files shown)


server/data/nomisClientBuilder.js
```diff
@@ -141,9 +141,9 @@ module.exports = token => {
       return nomisPut({ path, body })
     },
 
-    async putChecksPassed(bookingId) {
+    async putChecksPassed({ bookingId, passed }) {
```
romidane: would you want the `passed` to be defaulted?
i.e 
`{ passed = false }`
Or are you fine with `undefined` being a falsy?

server/routes/routeWorkers/standard.js
```diff
@@ -106,15 +106,15 @@ module.exports = ({ formConfig, licenceService, sectionName, nomisPushService, c
...
-      await nomisPushService.pushStatus(
-        bookingId,
-        {
-          type: formName,
-          status: !isEmpty(pushConfig.status) ? getIn(updatedLicence, pushConfig.status) : undefined,
-          reason: !isEmpty(pushConfig.reason) ? getIn(updatedLicence, pushConfig.reason) : undefined,
-        },
-        username
-      )
+      const status = !isEmpty(pushConfig.status) ? getIn(updatedLicence, pushConfig.status) : undefined
```
romidane: It like seems just doing 
`const status = getIn(updatedLicence, pushConfig.status)` 
would be enough as it would return the value or undefined?

atodd-noms: if pushConfig.status isn't defined then getIn will return the whole licence

server/data/nomisClientBuilder.js
atodd-noms: Good point. I'd rather a specific value was required because there's no obvious default behaviour. I'll change it so it's an error if no value supplied.

test/data/nomisClientTest.js
```diff
@@ -535,20 +552,40 @@ describe('nomisClient', () => {
...
         })
         .reply(200, { result: 'answer' })
 
-      return expect(nomisClient.putChecksPassed('aaa', { approvalStatus: 'status' })).to.eventually.eql({
+      return expect(nomisClient.putChecksPassed({ bookingId: 'aaa', passed: true })).to.eventually.eql({
         result: 'answer',
       })
     })
+
+    it('should throw error if passed is undefined', () => {
```
romidane: I think this can be done with 2 tests
```
it('should throw error if passed is undefined', () => {
```
and

```
it('should throw error if passed is not a boolean', () => {
```

I don't think it's necessary checking for the other types?

romidane: Arguably you probably only need 
```
it('should throw error if passed is not a boolean', () => {
```
What do you think?

atodd-noms: Agreed. I was trying to make it explicit and clear, but "not boolean" is pretty clear.

PR #867, opened 2019-04-25
Add DM final checks review screen

atodd-noms:


server/views/review/includes/booleanYesNo.pug
```diff
@@ -0,0 +1,8 @@
+mixin booleanYesNo(id, property)
+  span(id=id).block.bold.midMarginBottom
+    if property === true
+      | Yes
+    else if property === false
+      | No
+    else
+      | -
```
atodd-noms: NB the purpose of this is to prevent 'undefined' being treated as false
The value will have been set explicitly to false when user input means false, but will be undefined when not yet answered

test/routes/viewModels/taskListModelsTest.js
```diff
@@ -1727,6 +1727,52 @@ describe('TaskList models', () => {
   })
 
   describe('dmTasks', () => {
+    const eligibilitySummary = { task: 'eligibilitySummaryTask' }
```
atodd-noms: Mostly just factoring out a lot of long-winded duplication

PR #873, opened 2019-05-07
Meaningful error page when 409 Conflict on push-to-nomis

atodd-noms:


server/app.js
```diff
@@ -388,9 +388,8 @@ function handleKnownErrors(error, req, res, next) {
     logger.error(`Bad csurf token: ${error.stack}`)
   }
 
-  if (error.name === 'NoToken') {
-    logger.error('No token found for user')
-    return res.redirect('/logout')
+  if (error.name === 'NomisPushConflict') {
+    return res.render('nomis-push-error-409', { error })
   }
```
atodd-noms: This was never used - the NoToken error was deleted months ago and this left by accident

server/services/nomisPushService.js
```diff
@@ -15,15 +16,31 @@ module.exports = (nomisClientBuilder, signInService) => {
...
   }
 
   async function pushChecksPassed({ bookingId, passed, username }) {
     const systemTokens = await signInService.getClientCredentialsTokens(username)
     const nomisClient = nomisClientBuilder(systemTokens.token)
 
     logger.info(`Pushing checks passed=${passed} to nomis for bookingId: ${bookingId}`)
-    return nomisClient.putChecksPassed({ bookingId, passed })
+    try {
+      return await nomisClient.putChecksPassed({ bookingId, passed })
```
romidane: the is no point doing a `return await` you can just do 
```
return nomisClient.putChecksPassed({ bookingId, passed })
``` 
since the caller does the waiting..

But since you are handling the errors in this function rather then the caller I guess it's not a problem 👍

atodd-noms: yep, have to do it to catch the errors. Just return means the errors don't go in to the catch

PR #878, opened 2019-05-08
Approved Premises flow

atodd-noms:


server/routes/config/curfew.js
```diff
@@ -64,6 +64,14 @@ module.exports = {
...
     },
   },
+  approvedPremisesChoice: {
+    pageDataMap: ['licence'],
+    nextPath: {
+      discriminator: 'decision',
+      ApprovedPremises: '/hdc/curfew/approvedPremisesAddress/',
+      OptOut: '/hdc/taskList/',
+    },
+  },
```
atodd-noms: config for the standard route handler - this says where to go next based on the user's answer to the question

server/routes/curfew.js
```diff
@@ -1,15 +1,87 @@
...
+      return 'ApprovedPremises'
+    }
+
+    return null
+  }
+
+  function isYes(licence, pathSegments) {
+    const answer = getIn(licence, pathSegments)
+    return answer && answer === 'Yes'
+  }
```
atodd-noms: all the above is as per similar routes elsewhere - get the data out of the licence record and adapt for display

server/routes/curfew.js
```diff
@@ -1,15 +1,87 @@
...
+  const approvedPremisesContents = {
+    OptOut: { approvedPremises: { required: 'No' } },
+    ApprovedPremises: { approvedPremises: { required: 'Yes' } },
+  }
+
+  const proposedAddressContents = {
+    OptOut: { optOut: { decision: 'Yes' } },
+    ApprovedPremises: { optOut: { decision: 'No' } },
+  }
+
```
atodd-noms: Also similar to other routes - given the inputs, replace the related sections of licence data with the new values that match the user's choices.

Then if opted out, push to nomis.

Send to next view according to user's answer

PR #879, opened 2019-05-09
Where multiple relations, get all ids, ignore missing / select highest

atodd-noms:
Where there are two or more relationships of type ResponsibleOfficer assigned to an offender in NOMIS:

Previously we were using the one with the highest person ID and getting the delius ID for that person

However, the delius ID may not always exist, so now we lookup the delius ID for each person ID,

If there are still multiple delius IDs, pick the one with the highest person id

### DIFF
--- a/server/services/prisonerService.js
+++ b/server/services/prisonerService.js
@@ -70,8 +70,19 @@ function createPrisonerService(nomisClientBuilder) {
 
     try {
       const nomisClient = nomisClientBuilder(token)
-      const roRelation = await getRoRelation(nomisClient, bookingId)
-      return await getRoDetails(nomisClient, roRelation)
+      const roPersons = await getRoPersons(nomisClient, bookingId)
+
+      const allRoDetails = await Promise.all(
+        roPersons.map(roPerson => {
+          return getRoDetails(nomisClient, roPerson)
+        })
+      )
+
+      if (allRoDetails.length > 1) {
+        logger.warn(`Multiple RO relationships for bookingId: ${bookingId}`, allRoDetails)
+      }
+
+      return await selectRo(allRoDetails)
     } catch (error) {
       if (error instanceof RoRelationshipError) {
         logger.error(`findResponsibleOfficer for: ${bookingId}`, error.message)
@@ -83,7 +94,7 @@ function createPrisonerService(nomisClientBuilder) {
     }
   }
 
-  async function getRoRelation(nomisClient, bookingId) {
+  async function getRoPersons(nomisClient, bookingId) {
     const relations = await getRoRelations(nomisClient, bookingId)
     if (!relations[0]) {
       throw new RoRelationshipError('No RO relationship')
@@ -91,13 +102,11 @@ function createPrisonerService(nomisClientBuilder) {
 
     const personIds = relations.filter(r => !isEmpty(r.personId))
 
-    if (personIds.length > 1) {
-      logger.warn(`Multiple RO relationships for bookingId: ${bookingId}`)
-    } else if (isEmpty(personIds)) {
+    if (isEmpty(personIds)) {
       throw new RoRelationshipError('No RO person identifier')
     }
 
-    return personIds.sort((a, b) => b.personId - a.personId)[0]
+    return personIds
   }
 
   async function getRoRelations(nomisClient, bookingId) {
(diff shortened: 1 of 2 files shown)


server/services/prisonerService.js
```diff
@@ -83,21 +94,19 @@ function createPrisonerService(nomisClientBuilder) {
...
-    if (personIds.length > 1) {
-      logger.warn(`Multiple RO relationships for bookingId: ${bookingId}`)
-    } else if (isEmpty(personIds)) {
+    if (isEmpty(personIds)) {
       throw new RoRelationshipError('No RO person identifier')
     }
 
-    return personIds.sort((a, b) => b.personId - a.personId)[0]
+    return personIds
   }
```
atodd-noms: we were picking the highest person id, then getting the delius id for this person, but there may not always be a delius ID so we now carry on with all the person ids and pick on e later

server/services/prisonerService.js
```diff
@@ -70,8 +70,19 @@ function createPrisonerService(nomisClientBuilder) {
...
       const nomisClient = nomisClientBuilder(token)
-      const roRelation = await getRoRelation(nomisClient, bookingId)
-      return await getRoDetails(nomisClient, roRelation)
+      const roPersons = await getRoPersons(nomisClient, bookingId)
+
+      const allRoDetails = await Promise.all(
+        roPersons.map(roPerson => {
+          return getRoDetails(nomisClient, roPerson)
+        })
+      )
```
atodd-noms: we've now got multiple person ids and need to lookup the delius id for them all

server/services/prisonerService.js
```diff
@@ -114,25 +123,28 @@ function createPrisonerService(nomisClientBuilder) {
...
-  async function getRoDetails(nomisClient, roRelation) {
-    const { personId } = roRelation
+  async function getRoDetails(nomisClient, roPerson) {
+    const { personId } = roPerson
     const personIdentifiers = await getPersonIdentifiers(nomisClient, personId)
 
     if (!personIdentifiers || personIdentifiers.length < 1) {
-      throw new RoRelationshipError('No person identifiers')
+      logger.warn(`No person identifiers for RO person: ${roPerson}`)
+      return roPerson
```
atodd-noms: we're checking each person id now, so we don't thrown errors here because that would stop us checking the next ones

server/services/prisonerService.js
```diff
@@ -149,6 +161,16 @@ function createPrisonerService(nomisClientBuilder) {
...
+  async function selectRo(allRoDetails) {
+    const roWithDeliusIds = allRoDetails.filter(r => !isEmpty(r.deliusId))
+
+    if (isEmpty(roWithDeliusIds)) {
+      throw new RoRelationshipError('No RO with a Delius staff code')
+    }
+
+    return roWithDeliusIds.sort((a, b) => b.personId - a.personId)[0]
+  }
+
```
atodd-noms: discard results without a delius id, and from what's left sort by person id to get the most recent person

test/services/prisonerServiceTest.js
```diff
@@ -310,42 +338,53 @@ describe('prisonerDetailsService', () => {
...
       },
     ]
 
     sortingScenarios.forEach(scenario => {
       it(`should select person id where ${scenario.label}`, async () => {
         nomisClientMock.getRoRelations.resolves(scenario.relationships)
-        await service.getResponsibleOfficer('123', 'username')
-        expect(nomisClientMock.getPersonIdentifiers).to.be.calledWith(scenario.selected)
+
+        scenario.identifiers.forEach((response, index) => {
```
romidane: This is a bit complex can you explain what your testing here?

atodd-noms: This is stubbing different responses for each call, ie if there are 3 relationships it makes the next call 3 times and I need a different response for each call

atodd-noms: Would it be clearer to use stub.withArgs(blah).resolves(responseForBlah)?

atodd-noms: Re-written more explicitly

test/services/prisonerServiceTest.js
```diff
@@ -308,45 +336,62 @@ describe('prisonerDetailsService', () => {
...
+      nomisClientMock.getPersonIdentifiers.withArgs(1).resolves(person1)
+      nomisClientMock.getPersonIdentifiers.withArgs(2).resolves(person2)
+      nomisClientMock.getPersonIdentifiers.withArgs(3).resolves(person3)
+
+      const ro = await service.getResponsibleOfficer('123', 'username')
+
+      expect(ro.com.deliusId).to.eql('delius2')
+    })
+
+    it('should get the delius ID for the highest person ID when multiple persons with delius IDs', async () => {
```
romidane: Nice much clearer +:1

PR #880, opened 2019-05-09
Add nginx proxy buffer config

atodd-noms:
We get a 502 Gateway error in some circumstances. https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=156&modal=detail&selectedIssue=LIC-960

The nginx proxy in EB gives "upstream sent too big header while reading response header from upstream".

This may actually be a red herring and there's a different error going on behind it, but I need to try this first because nothing is wrong locally, only wrong when behind nginx in AWS. If there is another error behind, fixing this part should reveal it.

The only way to configure nginx in AWS is via .ebextensions, which requires deploying a zip. We do the same thing for the batchloader where we have to configure a bigger file upload size limit.

It means we won't be able to use the Jenkins deployment for licences and have to revert to deploying via AWS console like we used to. It may be possible to adapt the Jenkins workflow to deploy zips from S3 but I don't want to spend time on that until I know this fix works.

See eg https://stackoverflow.com/questions/23844761/upstream-sent-too-big-header-while-reading-response-header-from-upstream

### DIFF
--- a/bin/plant-beanstalk.js
+++ b/bin/plant-beanstalk.js
@@ -1,5 +1,6 @@
 const fs = require('fs')
 const path = require('path')
+const AdmZip = require('adm-zip')
 
 const version = process.argv[2]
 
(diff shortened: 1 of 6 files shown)


bin/plant-beanstalk.js
```diff
@@ -19,3 +20,8 @@ const dockerrun = {
 const output = JSON.stringify(dockerrun, null, 2)
 
 fs.writeFileSync(path.resolve(__dirname, '../Dockerrun.aws.json'), output)
+
+const zip = new AdmZip()
+zip.addLocalFile(`${__dirname}/../Dockerrun.aws.json`)
```
romidane: Are you not worried about any of this erroring?

atodd-noms: Only a little - it only happens during the circle CI build in order to generate the deployment artifacts, so if misconfigured the build would fail, but it's not going to be a runtime issue

PR #881, opened 2019-05-13
LIC-978 Form creation route and service

atodd-noms:
See https://dsdmoj.atlassian.net/wiki/spaces/LIC/pages/1478492565/Offender+Forms

Route and service to allow generating PDF forms from templates combined with offender data.

Will be wired into various pages for print on demand.

### DIFF
--- a/server/services/formService.js
+++ b/server/services/formService.js
@@ -1,0 +1,67 @@
+const moment = require('moment')
+const { getIn } = require('../utils/functionalHelpers')
+
+module.exports = function createFormService(pdfService, pdfFormatter) {
+  async function generatePdf(templateName, licence, prisoner) {
+    const dateNow = moment().format('DD/MM/YYYY')
+
+    const values = {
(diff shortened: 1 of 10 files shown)


server/services/formService.js
```diff
@@ -0,0 +1,67 @@
...
+    return combine(prisonerInfo, [['firstName'], ['middleName'], ['lastName']], ' ')
+  }
+
+  function getCurfewAddress(address) {
+    return address ? combine(address, [['addressLine1'], ['addressLine2'], ['addressTown'], ['postCode']], '\n') : ''
+  }
+
+  const noAddress = 'there is no suitable address for you to live at'
+  const noTime = 'there is not enough time before you’re due to be released'
+  // const offence = 'you committed an offence while you’ve been in prison'
```
atodd-noms: This is coming next as soon as there is a definitiion of when it applies

PR #882, opened 2019-05-14
Lic 844 part3

atodd-noms:
Unfortunately this part of the app is quite intricate. It's all about determining labels and buttons and links based on what kind of user you are and what answers have been given so far during the licence processing. Because this adds a 3rd type of address with a completely different set of rules, it gets quite detailed and affects a lot of files - caselist, tasklist, and all the computations of licence status.

### DIFF
--- a/server/routes/viewModels/taskLists/caTasks.js
+++ b/server/routes/viewModels/taskLists/caTasks.js
@@ -72,6 +72,7 @@ module.exports = {
       bassReferralNeeded,
       bassWithdrawn,
       bassAccepted,
+      approvedPremisesRequired,
       curfewAddressApproved,
       addressUnsuitable,
       optedOut,
@@ -81,6 +82,8 @@ module.exports = {
     const bassAreaChecked = bassAreaCheck === 'DONE'
     const bassExcluded = ['Unavailable', 'Unsuitable'].includes(bassAccepted)
     const bassChecksDone = bassReferralNeeded && bassAreaChecked && !bassWithdrawn && !bassExcluded
+
+    const validAddress = approvedPremisesRequired || curfewAddressApproved || bassChecksDone
 
     return [
       {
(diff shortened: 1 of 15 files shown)


server/utils/licenceStatusLabels.js
```diff
@@ -160,14 +163,13 @@ function roProcessingLabel(licenceStatus) {
     if (licenceStatus.decisions.bassAreaNotSuitable) {
       return status.bassAreaRejected
     }
-  } else if (licenceStatus.tasks.curfewAddressReview === taskStates.UNSTARTED) {
-    return status.notStarted
   }
```
atodd-noms: this was redundant

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -119,7 +123,7 @@ module.exports = {
           href: '/hdc/victim/victimLiaison/',
           text: 'View/Edit',
         },
-        visible: curfewAddressApproved || bassChecksDone,
+        visible: !optedOut && validAddress,
```
romidane: probably worth sticking this in a variable it's used in several places :/

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -253,7 +260,7 @@ module.exports = {
           href: '/hdc/victim/victimLiaison/',
           text: 'View/Edit',
         },
-        visible: eligible && (curfewAddressApproved || bassOfferMade),
+        visible: eligible && validAddress,
```
romidane: `eligible && validAddress` is used in several places probably worth having a variable?

server/routes/viewModels/taskLists/caTasks.js
atodd-noms: I chose to leave it that way because it was more expressive.

"If they're not opted out and they have a valid address"

Hard to think of a concise name for a variable to replace that.

Instead, although it means a small duplication (which is fine because it's easier to follow because the order is important and written out in full it's an exact reflection of the resulting view) it's probably clearer to pull out the small subset of tasks that ARE shown when opted out, and then the "opted in" case is cleaner.

server/routes/viewModels/taskLists/caTasks.js
atodd-noms: As above, have split out the small sub-set of tasks that apply when ineligible, then the conditions in the "eligble" task list are cleaner

PR #885, opened 2019-05-14
Lic 818 notification scheduler

atodd-noms:
See https://dsdmoj.atlassian.net/wiki/spaces/LIC/pages/1393918299/Notifications

This service runs some function (in this case sending reminder notifications to ROs) on a cron schedule.
The npm library node-schedule is used for cron-based scheduling.

The jobs are created at application startup. An admin screen is presented for internal admin users to enable/disable scheduled jobs. Cron schedules are set in config via an envvar (no requirement for user config at this stage)

To avoid multiple application instances in AWS all running the same scheduled job, a Postgres "advisory lock" database lock is used to ensure that only one instance gets to run the job.

To avoid scenarios where the job runs so quickly that it finishes before a second instance attempts to acquire the lock, a delay is introduced before releasing the lock.

### DIFF
--- a/server/services/jobSchedulerService.js
+++ b/server/services/jobSchedulerService.js
@@ -1,0 +1,87 @@
+const schedule = require('node-schedule')
+const advisoryLock = require('advisory-lock').default
+const moment = require('moment')
+const config = require('../config')
+const logger = require('../../log.js')
+const { onceOnly } = require('./jobs/jobUtils')
+
+module.exports = function createJobSchedulerService(notificationJobs) {
+  const connectionString = `postgres://${config.db.username}:${config.db.password}@${config.db.server}:5432/${
+    config.db.database
+  }`
+
+  const { active, overlapTimeout } = config.jobs
+
+  const roLock = advisoryLock(connectionString)('roReminders')
+
+  const jobs = [
+    {
+      name: 'roReminders',
+      spec: config.jobs.roReminders,
+      lock: roLock,
+      function: onceOnly(notificationJobs.roReminders, roLock, 'roReminders', overlapTimeout),
+    },
+  ]
+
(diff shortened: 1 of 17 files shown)


server/services/jobSchedulerService.js
```diff
@@ -0,0 +1,87 @@
+const schedule = require('node-schedule')
+const advisoryLock = require('advisory-lock').default
+const moment = require('moment')
+const config = require('../config')
+const logger = require('../../log.js')
+const { onceOnly } = require('./jobs/jobUtils')
+
+module.exports = function createJobSchedulerService(notificationJobs) {
+  const connectionString = `postgres://${config.db.username}:${config.db.password}@${config.db.server}:5432/${
```
romidane: Silly question but is the port guaranteed?

atodd-noms: It is (via docker locally, or terraform in AWS).

Still, pulled it out to common config to remove duplication

atodd-noms: Also cleaned up env var names and made it possible to start the app with all scheduled jobs off, then start them individually when needed

https://github.com/ministryofjustice/licences/pull/885/commits/62d22962eeaca9ac87816c52e31f1d112bad70c3

timharrison-moj: What happens if the env vars are incorrect or not supplied and the connection to Postgresql is not able to be made? Is there an alert / log / fatal error caught?

atodd-noms: It's the same as the app itself, so the healthchecks would have already failed and application startup wouldn't complete.

Having said that, the databse lock code should be separate from the scheduling code, and the error logging should be better so I'm making some changes.

PR #893, opened 2019-05-20
Pull logic out of view and fix section display when approved premises

atodd-noms:


server/routes/viewModels/reviewModels.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  getReviewSections,
+}
+
+function getReviewSections({ decisions }) {
+  if (!decisions) {
+    return {}
+  }
+
+  const isBass = decisions.bassReferralNeeded === true
```
atodd-noms: The values could be undefined, not just true/false, hence the explicit check

PR #896, opened 2019-05-22
Lic 986 additional forms

atodd-noms:


server/services/formService.js
```diff
@@ -32,6 +38,44 @@ module.exports = function createFormService(pdfService, pdfFormatter) {
     )
   }
 
+  function getDateValue(data, path) {
+    const value = getValue(data, path)
+    try {
+      if (moment(value, 'DD-MM-YYYY').isValid()) {
+        return moment(value, 'DD-MM-YYYY').format(formsDateFormat)
```
romidane: which is the method that throws an exceptions here?
`moment(value, 'DD-MM-YYYY').format(formsDateFormat)` ?
or 
`moment(value, 'DD-MM-YYYY').isValid()` ?

Because if it's not `moment(value, 'DD-MM-YYYY').isValid()` then why does it need to be in a try catch?

atodd-noms: Fixed

PR #897, opened 2019-05-23
Scheduled job runner with distributed locking

atodd-noms:
- page (route and view) for internal admin users (Jim) to activate/deactivate scheduled jobs & see last outcome
- config for job scheduling
- job scheduler using node-schedule to set nightly jobs with cron spec
- 'run once' utility using Postgres advisory locks so that multiple app instances in AWS don't cause duplicate job executions


server/config.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
     server: get('DB_SERVER', 'localhost'),
     database: get('DB_NAME', 'licences'),
     sslEnabled: get('DB_SSL_ENABLED', 'true'),
+    port: 5432,
```
timharrison-moj: Can this port number be overridden with an environment variable so that is could be deployed and configured in an environment where PG is running on a non-standard port?

atodd-noms: Done. Never has been, because it's dockerised so no reason for it not to be 5432

PR #900, opened 2019-05-28
Lic 818 notification scheduler config

atodd-noms:


atodd-noms:
It's documented in the wiki - https://dsdmoj.atlassian.net/wiki/spaces/LIC/pages/1393918299/Notifications#Notifications-ScheduledReminders

It's also temporary because it needs to be replaced with AWS cron

PR #917, opened 2019-06-13
Add delius API healthcheck

atodd-noms:


server/data/healthcheck.js
```diff
@@ -15,50 +16,34 @@ function dbCheck() {
...
+function pdfApiCheck() {
+  return serviceApiCheck('pdf', `${config.pdf.pdfServiceHost}/health`)
+}
 
-          if (result.status === 200) {
-            return resolve('OK')
-          }
+function nomisApiCheck() {
+  return serviceApiCheck('nomis', `${getHealthcheckUrl('nomis')}/health`)
+}
```
bell-pa: Is getHealthCheckUrl necessary? It looks as if it is only called with argument 'nomis', so perhaps this could be simplified. `${config.nomis.apiUrl.replace('api','')}/health` perhaps? but maybe that's too compressed.

PR #918, opened 2019-06-14
Lic 969 use delius prep refactoring

atodd-noms:
This looks bigger than it really is, but don't panic.

It's mostly just moving some code to a different place.

The caseListService and prisonerService had some code that talked to the nomis API to get the offenders for an RO (caseListService), or the RO for an offender (prisonerService)

To allow swapping in code that gets this info from delius/community API instead, I've extracted that code to a nomis-specific "ro service".

In another PR, there will be a delius variant of the ro service and a feature switch in the app setup to choose between nomis or delius.

All the code in nomisRoService is not new, it's just moved.

eg in caseListService instead of 
`const offenders = await nomisClient.getOffenderSentencesByBookingId(requiredIDs)`

we have
`const offenders = await roService.getROPrisoners(staffCode, token)`

Likewise, prisonerService instead of doing it internally just calls
`roService.findResponsibleOfficer(bookingId, token)`

and all the code for findResponsibleOfficer is moved into the new ro service.

Another change is to simplify the object structure for passing around the RO details.

For historic reasons the RO is still called COM (community offender manager) throughout the code, but the structure was unhelpfully nested so instead of using com.deliusId you had to do com.com.deliusId

I've got rid of that extra level and updated all the pug files etc that had references to com.com


server/index.js
```diff
@@ -28,13 +28,16 @@ const createNomisPushService = require('./services/nomisPushService')
...
 const createJobSchedulerService = require('./services/jobSchedulerService')
 const createNotificationJobs = require('./services/jobs/notificationJobs')
+const createNomisRoService = require('./services/nomisRoService')
 
 const signInService = createSignInService(audit)
 const licenceService = createLicenceService(licenceClient)
 const conditionsService = createConditionsService(config)
-const prisonerService = createPrisonerService(nomisClientBuilder)
+const nomisRoService = createNomisRoService(nomisClientBuilder)
+const roService = nomisRoService
```
atodd-noms: I realise this looks redundant but in the next PR this is where the feature switching will happen
eg
`const roService = config.roServiceType === 'DELIUS' ? deliusRoService : nomisRoService`

PR #920, opened 2019-06-14
Add Delius flavour of RO service, behind feature switch

atodd-noms:


server/data/deliusClient.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const { apiUrl } = config.delius
+
+module.exports = signInService => {
+  return {
+    getROPrisoners(deliusStaffCode) {
+      const path = `${apiUrl}/staff/staffCode/${deliusStaffCode}/managedOffenders`
+      const query = { current: true }
+      return deliusGet({ path, query })
```
timharrison-moj: The query parameter ?current=true is not required as this is added within the community proxy. Alternatively, I could rework the query parameters in the proxy?

server/data/deliusClient.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+
+    getResponsibleOfficer(offenderNo) {
+      const path = `${apiUrl}/offenders/nomsNumber/${offenderNo}/responsibleOfficers`
+      const query = { current: true }
```
timharrison-moj: Same as above - ?current=true not required.

test/data/deliusClientTest.js
```diff
@@ -0,0 +1,57 @@
...
+  describe('getROPrisoners', () => {
+    it('should return data from api', () => {
+      fakeDelius.get(`/staff/staffCode/1/managedOffenders?current=true`).reply(200, { key: 'value' })
+
+      return expect(deliusClient.getROPrisoners('1')).to.eventually.eql({ key: 'value' })
+    })
+
+    it('should reject if api fails', () => {
+      fakeDelius.get(`/staff/staffCode/1/managedOffenders?current=true`).reply(500)
+
```
timharrison-moj: Same thing - ?curren=true added by the proxy. Lets discuss Monday to see where the best place for this is?

server/data/deliusClient.js
atodd-noms: Done

PR #921, opened 2019-06-17
Add secondary reason for approval status push for serving recall

atodd-noms:
See also https://dsdmoj.atlassian.net/wiki/spaces/LIC/pages/1428783438/NOMIS+Push

https://dsdmoj.atlassian.net/browse/LIC-1003


atodd-noms:
Yep, perhaps not obvious in isolation but the second change was re-instating the test case. It's an example input + expected output that feeds into a test runner that isn't visible in the diff.

see
https://github.com/ministryofjustice/licences/pull/921/files#diff-00922d30f1a8773d036d6128abd26ec2R144

PR #922, opened 2019-06-17
Lic 1014 delius admin pages

atodd-noms:


server/views/admin/delius/responsibleOfficer.pug
```diff
@@ -0,0 +1,11 @@
+extends ../../layout
+
+block content
+
+  div.back-link-container.smallPaddingTop
+    a#back.link-back(href="/admin/delius") Back
+
+  div.pure-g.pure-u-1
```
timharrison-moj: Wonder if we need a few more details from the response here? There may be duplicates that we can filter based on either OM or OM status?

atodd-noms: Yeah, I think we'll probably add to this. Just wanted to get the first part up and running then we can review and design and specify what else is needed

PR #926, opened 2019-06-21
LIC-1013 "There is a problem" displayed in RO check details screen

bell-pa:
Fixes for problems described in LIC-1013.  They correct missing or incorrect error messages and missing text on the review/licenceDetails page.

### DIFF
--- a/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/pages/assessment/LicenceDetailsPage.groovy
+++ b/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/pages/assessment/LicenceDetailsPage.groovy
@@ -9,6 +9,7 @@ import uk.gov.justice.digital.hmpps.licences.modules.HeaderModule
 import uk.gov.justice.digital.hmpps.licences.modules.OffenderDetailsModule
 import uk.gov.justice.digital.hmpps.licences.modules.ReportingDetailsModule
 import uk.gov.justice.digital.hmpps.licences.modules.RiskDetailsModule
+import uk.gov.justice.digital.hmpps.licences.modules.AddressDetailsModule
 
 class LicenceDetailsPage extends Page {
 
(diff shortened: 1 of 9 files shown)


licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/pages/assessment/LicenceDetailsPage.groovy
```diff
@@ -35,9 +36,17 @@ class LicenceDetailsPage extends Page {
...
     }
-  }
 
+    proposedAddressError(required: false) { $('#proposed-address-error').text() }
+    riskError(required: false) { $('#risk-error').text() }
+    victimLiaisonError(required: false) { $('#victim-error').text() }
+    curfewHoursError(required: false) { $('#curfew-hours-error').text() }
+    additionalConditionsError(required: false) { $('#noConditionsError').text() }
+    reportingInstructionsError(required: false) { $('#reporting-error').text() }
+  }
```
atodd-noms: There is an alternative way of doing this that would be more flexible. Slightly less precise as a specification but less likely to be broken with changes to the page. Find the error messages by the class etc.

(even better if we used data-qa attributes to explicitly identify error messages)

eg add to the existing ErrorModule (which seems to be unused - I think there have been some merge errors in the past)

Many ways to do this eg select section or field validation errors separately:

```groovy
class ErrorModule extends Module {
  static content = {
    fields(required: false) {
      $('span.alert').collect { it.text() }
    }
    sections(required: false) {
      $('p.alert').collect { it.text() }
    }
  }
}
```
Then in the spec, check the expected error text.
```groovy
then: 'I see the error messages for each uncompleted section' 
  'Enter the risk management and victim liaison details' in errors.sections
  'Another error' in errors.sections
```
or errors.contains('expected') or whatever. 50 ways to do it in Groovy

It's not as specific because it doesn't check that the error message is in the right HTML id, but that's unlikely to be an issue and this way if the HTML id changes or more are added, it doesn't require changing the Geb page model.

The ErrorModule can be reused across pages, and then we'd have a consistent approach throughout the specs.

I'd even recommend a tech debt task to add data-qa attributes to validation error messages of different kinds, and add more feature specs like this one. I think we're a bit light on them.

PR #940, opened 2019-07-03
LIC-1027 DM name missing off the DM notification: 

bell-pa:
Favour the 'req.user.name' field over 'req.user.username' for sendingUserName

### DIFF
--- a/test/utils/userProfileTest.js
+++ b/test/utils/userProfileTest.js
@@ -1,0 +1,23 @@
+const { sendingUserName } = require('../../server/utils/userProfile')
+
+describe('sendingUserName', () => {
+  it('should extract username from user', () => {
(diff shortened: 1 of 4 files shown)


test/utils/userProfileTest.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(sendingUserName({ username: 'A', name: 'B' })).to.eql('B')
+  })
+
+  it('should fall back to username when name is not a string', () => {
+    expect(
+      sendingUserName({
+        username: 'A',
+        name: {
+          familyName: 'C',
+          givenName: 'D',
```
atodd-noms: if it was this couldn't we join givenName and familyName?

PR #943, opened 2019-07-05
Consistent casing, and fix a couple of layout bugs in the PDF

atodd-noms:


server/services/formService.js
```diff
@@ -106,7 +106,7 @@ module.exports = function createFormService(pdfFormatter, conditionsService, pri
     return labels[pickFirst(reasons)] || ''
   }
 
-  async function getCurfewAddressCheckData({ agencyLocationId, licence, isBass, isAP, bookingId, token }) {
+  async function getCurfewAddressCheckData({ agencyLocationId, licence, isBass, isAp, bookingId, token }) {
     const [conditions, prisoner, caMailboxes] = await Promise.all([
```
bell-pa: A test might help here...

atodd-noms: It already has tests, I've just changed the case

atodd-noms: The tests are in formsTest. 
```
    it('requests AP data when is AP', () => {
      const apLicence = { licence: { curfew: { approvedPremises: { required: 'Yes' } } }, stage: 'DECIDED' }
      licenceService.getLicence = sinon.stub().resolves(apLicence)
      app = createApp('roUser')

      return request(app)
        .get('/hdc/forms/curfewAddress/1')
        .expect(200)
        .expect('Content-Type', 'application/pdf')
        .expect(res => {
          expect(Buffer.isBuffer(res.body)).to.equal(true)
        })
        .expect(() => {
          expect(formService.getCurfewAddressCheckData).to.be.calledOnce()
          expect(formService.getCurfewAddressCheckData).to.be.calledWith({
            agencyLocationId: '123',
            licence: apLicence.licence,
            isBass: false,
            isAp: true,
            bookingId: '1',
            token: 'system-token',
          })
        })
    })
```
the test wasn't missing, it was an incorrect test because it didn't match the case in the service, because the service was using a casing that is inconsistent with the rest of the app.

It is valid for bassReferralNeeded or approvedPremisesRequired to be undefined. Forced it to never be undefined instead.

PR #945, opened 2019-07-11
DT-72: Updated Dependency Health endpoint call to /ping

SimonMitchellMOJ:


server/app.js
```diff
@@ -259,33 +259,25 @@ module.exports = function createApp({
     next()
   })
 
-  // Express Routing Configuration
+  const health = healthFactory(
+    config.nomis.apiUrl.replace('/api', ''),
```
JamieDraperUK: We need to fix the env variables to be the root URL like the other projects, then in code, we add either /api or /health

PR #946, opened 2019-07-12
LIC-1041 PDF documents not rendering in preprod environment

bell-pa:
Experimental tweaks that may fix the problem.  The only way to test this is to deploy to the pre-prod environment because the errors do not happen in any of the earlier environments (stage, local, local docker etc).

My current theory is that the pre-prod environment has a white-list of ip address for ingress:  The licences node application uses a headless Chrome browser to render PDF documents from HTML source.  This process is managed by a node module called puppet + a thin wrapper written by Alistair.  The HTML documents have links to a CSS file and font files that are served by the node app. Before this change the HTML documents prefixed the css and font file urls with the fully qualified public domain name of the licences instance.  eg https://licences-preprod.service.hmpps.dsd.io/public/forms-pdf.css etc.

This would work in the stage environment because there's no whitelist.  i think it fails in preprod because the whitelist does not include the licence application's public ip addresses.

This fix replaces the public domain prefix with http://127.0.0.1:3000 

The chrome browser should be able to read the css and font files without dns lookup and by making a local request that is unaffected by whitelists.

### DIFF
--- a/server/views/forms/forms-layout.pug
+++ b/server/views/forms/forms-layout.pug
@@ -3,8 +3,7 @@ include includes/formsHeader
 doctype html
 html(lang="en")
   head
-    link(href= domain + "/public/stylesheets/forms-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
-
+    link(href= "http://127.0.0.1:3000/public/stylesheets/forms-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
   body
 
     block content
--- a/server/views/licences/hdc_ap.pug
+++ b/server/views/licences/hdc_ap.pug
@@ -3,7 +3,7 @@ include includes/licenceHeader
 doctype html
 html(lang="en")
   head
-    link(href= domain + "/public/stylesheets/licences-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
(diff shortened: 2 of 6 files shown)


server/views/forms/forms-layout.pug
```diff
@@ -3,8 +3,7 @@ include includes/formsHeader
 doctype html
 html(lang="en")
   head
-    link(href= domain + "/public/stylesheets/forms-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
-
+    link(href= "http://127.0.0.1:3000/public/stylesheets/forms-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
```
timharrison-moj: Is the localhost reference going to work in any other environment except locally?

server/views/licences/hdc_ap.pug
```diff
@@ -3,7 +3,7 @@ include includes/licenceHeader
 doctype html
 html(lang="en")
   head
-    link(href= domain + "/public/stylesheets/licences-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
+    link(href= "http://127.0.0.1:3000/public/stylesheets/forms-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
```
timharrison-moj: As above

PR #951, opened 2019-07-23
LIC-1032 Lic-1031 

bell-pa:
Migrate licence templates and one licence vary templates to new PDF renderer

### DIFF
--- a/server/views/licences/hdc_ap_pss.pug
+++ b/server/views/licences/hdc_ap_pss.pug
@@ -1,0 +1,132 @@
+include includes/licenceMixins
+
+doctype html
+html(lang="en")
+  head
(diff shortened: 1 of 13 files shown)


server/views/licences/hdc_ap_pss.pug
```diff
@@ -0,0 +1,132 @@
+include includes/licenceMixins
+
+doctype html
+html(lang="en")
+  head
+    link(href= "http://127.0.0.1:3000/public/stylesheets/licences-pdf.css?" + version, media="print", rel="stylesheet", type="text/css")
```
yasinmustafa: Does the IP and port number have to be hard coded here? appreciate there is a lot of legacy code that does the same.

PR #965, opened 2019-08-07
Onboarding scripts, user and staff mapping load scripts and instructions.

timharrison-moj:
These were the scripts used to load the North East user data and will be useful for other regions to follow.

### DIFF
--- a/data-load/user-load.sh
+++ b/data-load/user-load.sh
@@ -4,7 +4,7 @@
 #
 # Parameters:
 #       1. ENV - [t3|t2|preprod|prod]
-#       2. CLIENT - the client Id and secret, comma-separated
+#       2. CLIENT - the client Id and secret, colon-separated 
 #       3. USER - the name of the client used to authenticate 
 #       4  BATCH - the size of batches (with 30 second pauses between them)
 #       5. FILE - the name of the file containing the user data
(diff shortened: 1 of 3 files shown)


data-load/user-load.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+#
+# Script to load a batch of new users into OAuth for new service access.
+#
+# Parameters:
+#       1. ENV - [t3|t2|preprod|prod]
+#       2. CLIENT - the client Id and secret, comma-separated
```
petergphillips: should be colon separated I think

data-load/user-load.sh
```diff
@@ -0,0 +1,100 @@
...
+  HOST="https://gateway.$ENV.nomis-api.service.hmpps.dsd.io"
+fi
+
+# Check whether the file exists and is readable
+if [[ ! -f "$FILE" ]]; then
+  echo "Unable to find file $FILE"
+  exit 1
+fi
+
+# Get  token for the client name / secret and store it in a hidden file .access_token
```
petergphillips: doesn't store in a hidden fie, we just use `jq` to extract out the `access_token` from the json

data-load/user-load.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+#
+# Script to load a batch of new users into OAuth for new service access.
+#
+# Parameters:
+#       1. ENV - [t3|t2|preprod|prod]
```
petergphillips: won't work for t2 out of the box as the `HOST` will be incorrect

data-load/user-load.sh
timharrison-moj: Good spot. Wrong words though does the right thing.

data-load/user-load.sh
timharrison-moj: Corrected to stored in env var TOKEN.

data-load/user-load.sh
timharrison-moj: Ok - I added the correct gateway address for using t2 so it would work there too.

PR #975, opened 2019-10-02
LIC-1085 Handling 404 errors when retrieving managed offenders. 

andrewrlee:
Also upgrading a couple of libraries

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -849,6 +849,11 @@
         "type-is": "~1.6.16"
       }
     },
+    "bowser": {
+      "version": "2.6.1",
+      "resolved": "https://registry.npmjs.org/bowser/-/bowser-2.6.1.tgz",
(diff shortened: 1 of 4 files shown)


andrewrlee:
> Looks good so far, but wondering whether all scenarios are covered here? The deliusROService should respond to the app in the same as as the nomisRoService for the requests - a) the list of offenders for a staff id, and b) the responsible officer for an offender, as the two services are substitutes for each other. There are probably two scenarios to look out for - 1) a 404 when the staff id or nomsId does not exist in Delius, and 2) a 200 OK but with no data for when the requested staffId/nomsId was found in Delius but had no offenders assigned or no responsible officer assigned. Provided the two services respond the same way to these requests it should be fine. Happy to approve this if you think it covers the scenarios above?

Pretty sure that’s ok - this handles the 404 and the api looks like it always return an empty list (rather than null). The code in licences copes with empty lists

The `findResponsibleOfficerByOffenderNo` copes with the officer not being found.

PR #980, opened 2019-10-14
LIC-1087 - Optional phone number in curfew address caused bug in task…

GurnankCheema:
…list page

The curfew address can optionally include a phone number. However in enabling this path,
the tasklist page became incorrect in that it still wanted the phone number to be included before it would let user continue.
Have removed this conditionality so that if user doesn't input a phone number (and has slected the checkbox for Main Occupier)
, the tasklist page's Curfew Address section  will show 'Completed' and a 'change' link.

### DIFF
--- a/server/utils/licenceStatus.js
+++ b/server/utils/licenceStatus.js
@@ -439,8 +439,15 @@ function getCurfewAddressState(licence, optedOut, bassReferralNeeded, curfewAddr
       return taskStates.STARTED
     }
 
-    const required = ['cautionedAgainstResident', 'addressLine1', 'addressTown', 'postCode', 'telephone']
+    const required = ['cautionedAgainstResident', 'addressLine1', 'addressTown', 'postCode']
+
     if (required.some(field => !address[field])) {
+      return taskStates.STARTED
+    }
+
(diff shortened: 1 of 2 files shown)


server/utils/licenceStatus.js
```diff
@@ -439,7 +439,7 @@ function getCurfewAddressState(licence, optedOut, bassReferralNeeded, curfewAddr
       return taskStates.STARTED
     }
 
-    const required = ['cautionedAgainstResident', 'addressLine1', 'addressTown', 'postCode', 'telephone']
+    const required = ['cautionedAgainstResident', 'addressLine1', 'addressTown', 'postCode']
```
andrewrlee: shouldn't this be conditional based on whether the offenderIsTheMainOccupier?

Also think you are missing a test to cover this?

GurnankCheema: The phone number is dependant on the main occupier checkbox but that is done in the Curfew Address page (LIC-1080) and not in the task list page.
I wrote unit tests that check for the existence/absence of the phone number as part of LIC-1080. 
This (task list) page doesn't actually refer to or display the phone number in any way.

GurnankCheema: new tests included

PR #986, opened 2019-10-25
Dcs 205 error in how licence condition is displaying

GurnankCheema:
A user was previously unable to record the issues the prisoner needs to address as part of his/her Licence Conditions. 

Have added a new set of checkboxes in the hdc/licenceConditions/additionalConditions page that define each of the Abuse and Behaviours.

Upon making a selection from these checkboxes and navigating to the hdc/licenceConditions/conditionsSummary page, the user can then see the selections they have made, inside the main body text of the page.

The selections made will also be displayed in the PDF of the licence given to the prisoner.

### DIFF
--- a/server/routes/forms.js
+++ b/server/routes/forms.js
@@ -21,19 +21,6 @@ module.exports = ({ formService }) => router => {
         licenceStatus,
       } = res.locals
 
-      const abuseAndBehaviours = getIn(res.locals.licence, [
-        'licence',
-        'licenceConditions',
-        'additional',
-        'COMPLYREQUIREMENTS',
-        'abuseAndBehaviours',
-      ])
-
-      if (Array.isArray(abuseAndBehaviours)) {
-        licence.licenceConditions.additional.COMPLYREQUIREMENTS.abuseAndBehaviours = abuseAndBehaviours.map(item => {
-          return ` ${item}`
-        })
-      }
       const isBass = getIn(licenceStatus, ['decisions', 'bassReferralNeeded']) === true
       const isAp = getIn(licenceStatus, ['decisions', 'approvedPremisesRequired']) === true
 
--- a/server/routes/forms.js
+++ b/server/routes/forms.js
@@ -21,6 +21,19 @@ module.exports = ({ formService }) => router => {
         licenceStatus,
       } = res.locals
 
+      const abuseAndBehaviours = getIn(res.locals.licence, [
+        'licence',
+        'licenceConditions',
+        'additional',
+        'COMPLYREQUIREMENTS',
+        'abuseAndBehaviours',
+      ])
+
+      if (Array.isArray(abuseAndBehaviours)) {
+        licence.licenceConditions.additional.COMPLYREQUIREMENTS.abuseAndBehaviours = abuseAndBehaviours.map(item => {
+          return ` ${item}`
+        })
+      }
       const isBass = getIn(licenceStatus, ['decisions', 'bassReferralNeeded']) === true
       const isAp = getIn(licenceStatus, ['decisions', 'approvedPremisesRequired']) === true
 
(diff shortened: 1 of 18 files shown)


server/routes/forms.js
```diff
@@ -21,6 +21,19 @@ module.exports = ({ formService }) => router => {
...
+        'additional',
+        'COMPLYREQUIREMENTS',
+        'abuseAndBehaviours',
+      ])
+
+      if (Array.isArray(abuseAndBehaviours)) {
+        licence.licenceConditions.additional.COMPLYREQUIREMENTS.abuseAndBehaviours = abuseAndBehaviours.map(item => {
+          return ` ${item}`
+        })
+      }
```
bell-pa: I don't think this is the right place to format a specific data item for printing.
Line 40 below, calls the confusingly named getCurfewAddressCheckData function.  It is badly named because it does a *lot* more than retrieve curfew address check data.  This function formats all the licence data for the address check form.
Formatting the licence conditions is delegated to conditionsService.js (getFullTextForApprovedConditions).
That is where this piece of formatting should go.  Probably in the function 'getConditionText', which looks surprisingly like the other small formatting function used by the licence PDF formatting code.  A similar tweak to this function should do the trick.  If it turns out that the changes are identical to those on the other function then I'd suggest extracting that code in to a separate module to remove the duplication.

GurnankCheema: Thanks. I'll have another go

server/views/licenceConditions/includes/appointmentName.pug
```diff
@@ -6,13 +6,13 @@ div.pure-g.paddingBottom
...
 
   div.multiple-choice
-    input(id='psychologist' type='radio' name='appointmentProfession' value='psychologist' checked=condition.user_submission && condition.user_submission.appointmentProfession === 'Psychologist')
+    input(id='psychologist' type='radio' name='appointmentProfession' value='psychologist' checked=condition.user_submission && condition.user_submission.appointmentProfession === 'psychologist')
     label(for='psychologist') Psychologist
 
   div.multiple-choice
-    input(id='medicalPractitioner' type='radio' name='appointmentProfession' value='medical practitioner' checked=condition.user_submission && condition.user_submission.appointmentProfession === 'Medical practitioner')
+    input(id='medicalPractitioner' type='radio' name='appointmentProfession' value='medical practitioner' checked=condition.user_submission && condition.user_submission.appointmentProfession === 'medical practitioner')
     label(for='medicalPractitioner') Medical practitioner
```
bell-pa: Looks as if this part of the form has no test coverage.  Perhaps we need to add some tests.   I think this is outside the scope of this story/bug.  Consider raising a separate ticket for this.

server/views/review/includes/additionalConditions.pug
```diff
@@ -26,7 +26,10 @@ mixin additionalConditions(editable)
             if item.text
               | !{item.text}
             if item.variable
-              span.bold #{item.variable}
+              if (Array.isArray(item.variable))
+                span.bold #{item.variable.join(', ')}
+              if (typeof(item.variable) === 'string')
+                span.bold #{item.variable}
             if item.error
```
bell-pa: This kind of logic shouldn't be necessary here.

server/views/licenceConditions/includes/appointmentName.pug
GurnankCheema: ok

server/views/review/includes/additionalConditions.pug
GurnankCheema: i'll remove it and see what it does

server/routes/forms.js
GurnankCheema: The logic above (lines 24 to 36) was included to insert a space between the Behaviours. Have removed it and changed populateLicenceWithConditions() inside server/services/conditionsService. getFullTextForApprovedConditions  so that it places a space before each condition

PR #988, opened 2019-10-30
DCS-181 HDC - Risk Management Information not displaying for the Decision Maker

bell-pa:
When a curfew address is rejected then show the DM the Risk Management task iff the address was rejected by this task, not the 'Proposed curfew address' task.

### DIFF
--- a/server/routes/viewModels/taskLists/dmTasks.js
+++ b/server/routes/viewModels/taskLists/dmTasks.js
@@ -13,6 +13,7 @@ module.exports = ({ decisions, tasks, stage }) => {
   const {
     insufficientTimeStop,
     addressWithdrawn,
+    addressUnsuitable,
     curfewAddressRejected,
     bassReferralNeeded,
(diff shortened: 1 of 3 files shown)


server/routes/viewModels/taskLists/dmTasks.js
```diff
@@ -35,7 +36,7 @@ module.exports = ({ decisions, tasks, stage }) => {
   }
 
   if (addressWithdrawn || curfewAddressRejected) {
-    return [
+    const t = [
```
andrewrlee: is this supposed to be `tasks`? (a list of tasks)

bell-pa: Yes, but 'tasks' is already in scope.  Could be 'rejectedAddressTasks' I suppose. At the time I couldn't think of a suitable name and then forgot about it.  I'll fix in my next PR.

PR #1015, opened 2019-11-22
DCS-233 Extracting out service to retrieve RO contact details

andrewrlee:
This is to start to provide a central place where we can start to using delius and probation teams service as the source of org emails.

This will initially only be a fallback for when we haven't got the RO user's details in the staff ids table.

The naming might be a bit off here - going to talk to Louise about naming around RO vs COM, but RO is currently how we describe the role across licences

### DIFF
--- a/server/services/notificationService.js
+++ b/server/services/notificationService.js
@@ -7,7 +7,7 @@ const { getRoCaseDueDate, getRoNewCaseDueDate } = require('../utils/dueDates')
 
 module.exports = function createNotificationService(
   prisonerService,
-  userAdminService,
+  roContactDetailsService,
   configClient,
   notifyClient,
   audit,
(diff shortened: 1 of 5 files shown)


server/services/notificationService.js
```diff
@@ -131,12 +125,16 @@ module.exports = function createNotificationService(
       return []
     }
 
-    const [establishment, ro] = await Promise.all([
+    const [establishment, roContactInfo] = await Promise.all([
       prisonerService.getEstablishmentForPrisoner(bookingId, token),
-      userAdminService.getRoUserByDeliusId(deliusId),
+      roContactDetailsService.getRoContactInfo(deliusId),
```
brightonsbox: `roContactDetailsService.getContactDetails(deliusId)`?

PR #1018, opened 2019-11-26
DCS-248 split out ro and ca/dm notification sending

andrewrlee:
We want to have a separate workflow for sending notifications to RO users so breaking up the notification service into 3 separate components:
* RoNotifcationSender
* CaAndDmNotifcationSender
* NotificationSender (which the previous modules call)

Also then pulled down logic from the send route into a new notification service - this logic is only really relevant for ca/dm notifications and will change later.


test/services/notifications/notificationSenderTest.js
```diff
@@ -0,0 +1,216 @@
...
+  describe('notify', () => {
+    it('should do nothing if template ID not in active template list', async () => {
+      const notifications = [{ email: ['email@email.com'] }]
+      await service.notify({
+        sendingUserName: 'username',
+        bookingId: 123,
+        notifications,
+      })
+      expect(notifyClient.sendEmail).not.to.be.calledOnce()
+    })
```
bell-pa: Would the test pass if sendEmail was called twice? Perhaps 
`expect(notifyClient.sendEmail).not.to.be.called()`
 would be better.

andrewrlee: probably copied this from somewhere - will fix in a follow up,

PR #1021, opened 2019-12-01
DCS-248 Adding type checking

andrewrlee:
Add type checking on commit and fixing associated type errors.

There were lots of places where incorrect or unused parameters were being passed.
Also removed unused health status tracking (as this is monitored externally).

This encouraged being a lot more explicit about when things should be null.

It also highlighted some strange defensive coding, e.g: on a `restclient.get` method which allowed a path to not be specified rather than failing fast: 
`return async ({ path, query = '', headers = {}, responseType = '' } = {})`  
rather than 
`return async ({ path, query = '', headers = {}, responseType = '' })` 

Also adds npm task to perform checks:
`npm run typecheck`


server/routes/routeWorkers/standard.js
```diff
@@ -1,7 +1,7 @@
 const { getIn, isEmpty, firstItem, lastItem } = require('../../utils/functionalHelpers')
 const { getPathFor } = require('../../utils/routes')
 
-module.exports = ({ formConfig, licenceService, sectionName, nomisPushService, config }) => {
+module.exports = ({ formConfig, licenceService, sectionName, nomisPushService = null, config = null }) => {
```
andrewrlee: this component is created for each route, the last two parameters are optional depending on whether the routes use nomis push or not - seems like this could be tidied up a bit.

server.js
```diff
@@ -1,29 +1,5 @@
...
-
-function recordHealthResult(err, results) {
-  if (err) {
-    logger.error('healthcheck failed', err)
-    return
-  }
-  logger.info('healthcheck', results)
-  if (results.healthy && appInsights) {
-    appInsights.defaultClient.trackEvent('healthy', flattenMeta(results))
-  }
```
andrewrlee: when enabled, this polls and write health status events to app insights but this isn't enabled anywhere.
Removing this as if we required health status tracking we could implement a solution globally rather than embed in each app

server/data/healthcheck.js
```diff
@@ -1,4 +1,5 @@
 const superagent = require('superagent')
+/** @type {any} */
```
andrewrlee: There's a bug in the typescript definitions here - it doesn't recognise Agent as a constructor: https://github.com/node-modules/agentkeepalive/issues/76

server/services/healthcheck.js
```diff
@@ -12,6 +12,7 @@ const gatherCheckInfo = (total, currentValue) => Object.assign({}, total, { [cur
 
 const getBuild = () => {
   try {
+    // @ts-ignore
```
andrewrlee: ignore as file doesn't exist

server/services/deadlineService.js
```diff
@@ -28,7 +28,7 @@ module.exports = function createDeadlineService(licenceClient) {
   function rejectIfUnknown(role) {
     if (!roleStages[role] || !dueDateMethods[role]) {
       const message = `Unmatched role code for getting due dates: ${role}`
-      logger.error()
+      logger.error(message)
```
andrewrlee: little bug which was picked up

server/utils/errors.js
```diff
@@ -4,12 +4,14 @@ module.exports = {
 }
 
 function unauthorisedError() {
+  /** @type {any} */
```
andrewrlee: Can't set arbitrary properties on known types - real way to fix would be to define a type which extends `Error` and adds specific fields but leaving for now

tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "suppressImplicitAnyIndexErrors": true,
+    "skipLibCheck": true,
+    "noEmit": true,
```
andrewrlee: Just perform checks - don't actually produce `js`

log.js
```diff
@@ -72,5 +72,6 @@ if (appInsights) {
     return flattenMeta(meta)
   })
 }
-
-module.exports = logger
+/** @type {winston.LoggerInstance & {audit?: Function }} */
```
andrewrlee: Winston adds additional functions for each logger level defined which can't be covered by the ts definition. 
This defines an [intersection type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types) which is both a winston logger and one of these anonymous types which may have an audit function.

This allows us to call `logger.audit(message)` in `audit.js`

PR #1023, opened 2019-12-02
DCS-263: new migration file to create LDU table

GurnankCheema:
New table created to hold the Local Delivery Unit codes.
The presence of a code means that particular Local Delivery Unit is 'active'

### DIFF
--- a/server/data/activeLduClient.js
+++ b/server/data/activeLduClient.js
@@ -3,21 +3,20 @@ const db = require('./dataAccess/db')
 module.exports = {
   async addLdu(lduCode) {
     const query = {
-      text: `INSERT INTO active_local_delivery_units (ldu_code) VALUES ($1) ON CONFLICT (ldu_code) DO UPDATE SET ldu_code=active_local_delivery_units.ldu_code RETURNING ldu_code`,
+      text: `INSERT INTO active_local_delivery_units (ldu_code) VALUES ($1) 
+      ON CONFLICT (ldu_code) DO NOTHING`,
       values: [lduCode],
     }
-    const response = await db.query(query)
-    return response
+    return db.query(query)
   },
 
-  async doesLduExist(lduCode) {
+  async isLduPresent(lduCode) {
     const query = {
-      text: `SELECT ldu_code FROM active_local_delivery_units WHERE ldu_code=$1`,
+      text: `SELECT count(*) FROM active_local_delivery_units WHERE ldu_code=$1`,
       values: [lduCode],
     }
 
-    const response = await db.query(query)
-    const active = response && response.rows
-    return active ? 'true' : 'false'
+    const { rows } = await db.query(query)
+    return parseInt(rows[0].count, 10) > 0
   },
 }
--- a/server/data/activeLduClient.js
+++ b/server/data/activeLduClient.js
@@ -1,0 +1,23 @@
+const db = require('./dataAccess/db')
+
+module.exports = {
+  async addLdu(lduCode) {
+    const query = {
+      text: `INSERT INTO active_local_delivery_units (ldu_code) VALUES ($1) ON CONFLICT (ldu_code) DO UPDATE SET ldu_code=active_local_delivery_units.ldu_code RETURNING ldu_code`,
+      values: [lduCode],
+    }
+    const response = await db.query(query)
+    return response
+  },
+
+  async doesLduExist(lduCode) {
+    const query = {
+      text: `SELECT ldu_code FROM active_local_delivery_units WHERE ldu_code=$1`,
+      values: [lduCode],
+    }
+
+    const response = await db.query(query)
+    const active = response && response.rows
+    return active ? 'true' : 'false'
+  },
+}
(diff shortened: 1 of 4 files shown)


server/data/activeLduClient.js
```diff
@@ -0,0 +1,23 @@
...
+      values: [lduCode],
+    }
+    const response = await db.query(query)
+    return response
+  },
+
+  async doesLduExist(lduCode) {
+    const query = {
+      text: `SELECT ldu_code FROM active_local_delivery_units WHERE ldu_code=$1`,
+      values: [lduCode],
```
andrewrlee: In use of force we do something like this: 
```const getNumberOfPendingStatements = async (reportId, client = nonTransactionalClient) => {
  const { rows } = await client.query({
    text: `select count(*) from statement where report_id = $1 AND statement_status = $2`,
    values: [reportId, StatementStatus.PENDING.value],
  })
  return parseInt(rows[0].count, 10)
}```
by using count we can always assume that a row will exist.

File: server/data/activeLduClient.js [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    const response = await db.query(query)
+    const active = response && response.rows
+    return active ? 'true' : 'false'
```
andrewrlee: we would want to return a boolean here not string. 
Returning strings is confusing as 'false' is truthy

server/data/activeLduClient.js
```diff
@@ -0,0 +1,23 @@
+const db = require('./dataAccess/db')
+
+module.exports = {
+  async addLdu(lduCode) {
+    const query = {
+      text: `INSERT INTO active_local_delivery_units (ldu_code) VALUES ($1) ON CONFLICT (ldu_code) DO UPDATE SET ldu_code=active_local_delivery_units.ldu_code RETURNING ldu_code`,
+      values: [lduCode],
```
andrewrlee: probably don't want to do an update on conflict, would be tempted to `ON CONFLICT DO NOTHING`.

server/data/activeLduClient.js
andrewrlee: There is a general convention to name methods which return boolean `is*` - so in this case could call it something like `isLduPresent`

test/data/activeLduClientTest.js
```diff
@@ -0,0 +1,52 @@
...
+        query,
+      },
+    })
+  }
+
+  describe('doesLduExist', () => {
+    it('should call query once', () => {
+      queryStub = sinon.stub().resolves()
+      lduProxy().doesLduExist(lduCode)
+      expect(queryStub).to.have.callCount(1)
```
andrewrlee: these tests aren't really testing anything, would be worth looking at some of the other examples.

e.g, ones which check the statement and provided args: 
```
    const expectedParameters = ['NEW_STAGE', 'ABC123']

      const result = licencesProxy().updateStage('ABC123', 'NEW_STAGE')

      return result.then(() => {
        const { values } = queryStub.getCalls()[0].args[0]
        expect(values).to.eql(expectedParameters)
      })
    })
```

andrewrlee: or ```const expectedClause =
        'insert into licences (booking_id, licence, stage, version, vary_version) values ($1, $2, $3, $4, $5)'

      const result = licencesProxy().createLicence('ABC123')

      return result.then(() => {
        expect(queryStub.getCalls()[0].args[0].text).includes(expectedClause)
      })```

(there are better examples in use-of-force)

server/data/activeLduClient.js
GurnankCheema: changed to isLduPresent

GurnankCheema: I'm not 100% sure about the other point - using count, because it attempts to access [0] when there may be no rows at all. 
I'm thinking this might be better
    **const {rows} = await db.query(query)
    return getIn(rows, [[0]]) ? 'true' : 'false'**

server/data/activeLduClient.js
GurnankCheema: I did initially use booleans for true and false but the linter gave me the following error and that's why I changed to a string
  **21:33  error  Unnecessary use of boolean literals in conditional expression** 
I could disable linter for that line?

server/data/activeLduClient.js
GurnankCheema: I did consider doing ON CONFLICT DO NOTHING but then came across this article that says that nothing will be returned on a conflict. https://intellipaat.com/community/14280/how-to-use-returning-with-on-conflict-in-postgresql

So I used DO UPDATE just so we have a complete object on return

server/data/activeLduClient.js
GurnankCheema: done

test/data/activeLduClientTest.js
GurnankCheema: have revised tests using test/data/licenceClientTest.js as examples

server/data/activeLduClient.js
GurnankCheema: done

PR #1033, opened 2019-12-09
DCS-276: Create caService

GurnankCheema:
New service to get the LDU associated to an RO.
The RO to LDU relationship is defined in Delius. Delius is passed the bookingId in order to begin the search.


server/services/caService.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {DeliusClient} deliusClient
+ * @returns {CaService} caService
+ */
+
+const logger = require('../../log.js')
+const roService = require('./roService')
+
+module.exports = function createCaService(deliusClient, nomisClientBuilder) {
+  async function getLduForRo(bookingId, token) {
+    const ro = await roService(deliusClient, nomisClientBuilder).findResponsibleOfficer(bookingId, token)
```
andrewrlee: Services should only be created once and live the entire lifecycle of the application - not created per request. 
You can pass the created service in to the `createCaService`

server/services/caService.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @param {DeliusClient} deliusClient
+ * @returns {CaService} caService
+ */
+
+const logger = require('../../log.js')
+const roService = require('./roService')
+
+module.exports = function createCaService(deliusClient, nomisClientBuilder) {
```
andrewrlee: Can you add type signatures here?

server/services/caService.js
```diff
@@ -0,0 +1,30 @@
...
+  async function getLduForRo(bookingId, token) {
+    const ro = await roService(deliusClient, nomisClientBuilder).findResponsibleOfficer(bookingId, token)
+
+    if (ro.lduCode) {
+      logger.info(`Local delivery unit code received from Delius: ${ro.lduCode}`)
+    } else if (ro.message && ro.message.match(/Offender has not been assigned a COM:/)) {
+      logger.info('Offender has not been assigned a COM')
+    }
+
+    return ro.lduCode || ro.message
```
andrewrlee: there would be no way of being able to distinguish between these two results? They would both return a string - how you you know if its a valid ldu code?

andrewrlee: You'll also need to distinguish whether the RO is allocated or not: 

```
Call ROService.findResponsibleOfficerByBookingId
If error then return error
Call lduActiveClient with result.ldu_code 
Build a list of errors:
Error if not active
Error if !result.isAllocated
```

server/services/caService.js
GurnankCheema: I used the example of roService in creating the service. 
...

The caService was created in server/index.js and passed to createApp

So do you mean I shouldn't wrap the getLduForRo in createCaService, rather just export it on it's own, like this...
module.exports = async function getLduForRo(bookingId, token) {
const { lduCode, message } = 
await roService(deliusClient,nomisClientBuilder)
.findResponsibleOfficer( bookingId, token  )
 etc, etc


?

andrewrlee: this comment is more about only passing `roService` rather than creating it in the method: 

`function createCaService(roService) {`

server/index.js
```diff
@@ -38,13 +38,15 @@ const createNotificationJobs = require('./services/jobs/notificationJobs')
...
 const createRoService = require('./services/roService')
+const createCaService = require('./services/caService')
 
 const signInService = createSignInService()
 const licenceService = createLicenceService(licenceClient)
 const conditionsService = createConditionsService(config)
 const deliusClient = createDeliusClient(signInService)
 const probationTeamsClient = createProbationTeamsClient(signInService)
 const roService = createRoService(deliusClient, nomisClientBuilder)
+const caService = createCaService(deliusClient, nomisClientBuilder)
```
andrewrlee: so you would pass through `roService` here

server/services/caService.js
andrewrlee:
```/***
 * @param {RoService} roService
 * @returns {CaService} caService
 */
module.exports = function createCaService(deliusClient, nomisClientBuilder) {```

File: server/services/caService.js (hunk 1 above)
Comment (GurnankCheema) [reply server/services/caService.js]: oh, I see

File: server/index.js (hunk 4 above)
Comment (GurnankCheema) [reply server/index.js]: ok

File: server/services/caService.js (hunk 1 above)
Comment (GurnankCheema) [reply server/services/caService.js]: doh!

File: server/index.js (hunk 4 above)
Comment (GurnankCheema) [reply server/index.js]: ok

File: server/services/caService.js (hunk 2 above)
Comment (GurnankCheema) [reply server/services/caService.js]: ok

File: server/services/caService.js (hunk 3 above)
Comment (GurnankCheema) [reply server/services/caService.js]: ok

File: server/services/roService.js [hunk 5]
```diff
@@ -3,10 +3,12 @@
  * @typedef {import("../../types/licences").RoService} RoService
  * @typedef {import("../../types/delius").CommunityOrPrisonOffenderManager} CommunityOrPrisonOffenderManager
  * @typedef {import("../../types/licences").ResponsibleOfficerResult} ResponsibleOfficerResult
+ *  * @typedef {import("../../types/licences").ResponsibleOfficer} ResponsibleOfficer
```
andrewrlee: unnecessary `*`

test/services/caServiceTest.js
```diff
@@ -0,0 +1,74 @@
...
+
+    beforeEach(() => {
+      roService = {
+        findResponsibleOfficer: sinon.stub().resolves(responsibleOfficerOrError),
+      }
+    })
+
+    describe('getReasonForNotContinuing ', () => {
+      it('should return null because Ro is COM and ldu is active', async () => {
+        lduActiveClient.isLduPresent = sinon.stub().resolves(1)
```
andrewrlee: these should `isLduPresent ` be returning true or false.

Would be good to move  `caService = createCaService(roService, lduActiveClient)` into the `beforeEach`

test/services/caServiceTest.js
```diff
@@ -0,0 +1,74 @@
+const createCaService = require('../../server/services/caService')
+
+describe('caService', () => {
+  let roService
+  let caService
+  const lduActiveClient = { isLduPresent: () => {} }
+
+  describe('Responsible Officer is allocated and Ldu determined', () => {
+    const responsibleOfficerOrError = {
```
andrewrlee: maybe just call this `responsibleOfficer`? as it always is in this set of tests?

test/services/caServiceTest.js
```diff
@@ -0,0 +1,74 @@
...
+        responsibleOfficerOrError.isAllocated = false
+        const result = await caService.getReasonForNotContinuing('bookingId-1', 'token-1')
+        expect(result).to.eql('COM_NOT_ALLOCATED')
+        responsibleOfficerOrError.isAllocated = true
+      })
+    })
+  })
+
+  describe('Responsible officer not assigned or Ldu not determined', () => {
+    const responsibleOfficerOrError = {
```
andrewrlee: and maybe just always call this `Error`, as it always is, in this set of tests

server/services/roService.js
GurnankCheema: done

test/services/caServiceTest.js
GurnankCheema: done
The only reason I duplicated them was, I thought it needed to come after the respective 
lduActiveClient.isLduPresent stubs, rather than before

test/services/caServiceTest.js
GurnankCheema: done

test/services/caServiceTest.js
GurnankCheema: done

test/services/caServiceTest.js
andrewrlee: that's understandable but JS is 'pass by reference'. The lduActiveClient variable that you are manipulating is acting on the variable passed into the service (they both reference the same object in memory)

PR #1035, opened 2019-12-09
DCS-261 Fixing error handling and adding type info for prisoner service

andrewrlee:


server/services/reminderService.js
```diff
@@ -48,14 +67,29 @@ module.exports = function createReminderService(prisonerService, deadlineService
...
+        if (isError(result)) {
+          logger.error('Error loading data for reminder', result)
+        }
+        return result
+      })
+      .catch(error => {
+        logger.error('Error loading data for reminder', error.stack)
+        return { message: `Error loading data for reminder: ${error.message}` }
+      })
   }
```
bell-pa: If you're not going to use await then you don't need async.
Alternatively, replace the '.then' chain with await + try/catch

andrewrlee: I've replaced the `then` and `catch` with `await` and its lots clearer. Also meant I could get rid of the quite poor, additional methods in functionalHelpers

PR #1037, opened 2019-12-12
DCS-280 Extend staff id look up to call delius for missing users.

bell-pa:
Integration tests must be run against an updated licences-demo-mock service to succeed.

### DIFF
--- a/server/data/nomisClientBuilder.js
+++ b/server/data/nomisClientBuilder.js
@@ -1,3 +1,9 @@
+/**
+ * @typedef {import("../../types/elite2api").Role} Role
+ * @typedef {import("../../types/elite2api").Profile} Profile
+ *
+ */
+
 const moment = require('moment')
 const superagent = require('superagent')
 const logger = require('../../log')
@@ -96,11 +102,17 @@ module.exports = token => {
       return nomisGet({ path })
     },
 
+    /**
+     * @returns {Promise<Profile>}
+     */
     getLoggedInUserInfo() {
       const path = `${authUrl}/api/user/me`
       return nomisGet({ path })
     },
 
+    /**
+     * @returns {Promise<[Role]>}
+     */
     getUserRoles() {
       const path = `${authUrl}/api/user/me/roles`
       return nomisGet({ path })
--- a/server/services/caseListService.js
+++ b/server/services/caseListService.js
@@ -1,8 +1,21 @@
 /**
- * @typedef {import("../services/roService").RoService} RoService
+ * @template S, E
+ * @typedef {import("../utils/ResultTypes").Result<S, E>} Result
  */
+const R = require('ramda')
+const { Success, Fail } = require('../utils/Result')
+/**
+ * @typedef {import("../services/roService").RoService} RoService
+ * @typedef {import("../../types/delius").StaffDetails} StaffDetails
+ */
+
+/**
+ * @typedef DeliusId
+ * @property {string} staff_id
+ */
+
 const logger = require('../../log.js')
-const { isEmpty, getIn } = require('../utils/functionalHelpers')
+const { isEmpty } = require('../utils/functionalHelpers')
 
 /**
  * @param {RoService} roService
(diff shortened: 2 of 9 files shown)


server/services/caseListService.js
```diff
@@ -1,8 +1,16 @@
...
+const Result = require('../utils/Result')
 /**
  * @typedef {import("../services/roService").RoService} RoService
  */
 const logger = require('../../log.js')
-const { isEmpty, getIn } = require('../utils/functionalHelpers')
+const { isEmpty } = require('../utils/functionalHelpers')
+
+/**
+ * @typedef {Object<string[], string>} CaselistResponse
```
andrewrlee: didn't think you could have non string keys in js?

server/data/nomisClientBuilder.js
```diff
@@ -96,11 +102,17 @@ module.exports = token => {
       return nomisGet({ path })
     },
 
+    /**
+     * @returns {Promise<Profile>}
+     */
```
andrewrlee: does this type info get used anywhere without defining an interface?

server/services/caseListService.js
```diff
@@ -25,40 +33,74 @@ module.exports = function createCaseListService(nomisClientBuilder, roService, l
...
-      }
+  /**
+   * @param {string} username
+   * @returns {Promise<Result<string, string>>}
+   */
+  const getStaffCodeFromDelius = async username => {
+    const staffDetailsResult = await getStaffDetailsFromDelius(username)
+
+    return staffDetailsResult.flatMap(staffDetails => {
+      const staffCode = R.prop('staffCode', staffDetails)
```
andrewrlee: would avoid using `R.prop` as that returns `any` and suggests that staffDetails could potentially be null when it can't be here?

server/services/caseListService.js
```diff
@@ -25,40 +33,74 @@ module.exports = function createCaseListService(nomisClientBuilder, roService, l
...
+   * Assume username is assigned to an RO.  Look up this user in the local db (staff_ids table) and take the staff code.
+   * Alternatively if a unique staff code cannot be found in this db, ask Delius for a staff code.
+   *
+   * The code below is an ( ugly bodge / elegant solution ) that breaks the type contract on Result.orElse():
+   * staffCodeFromDB is a Result<string,string>, but getStaffCodeFromDelius() returns a Promise<Result<string,string>>
+   * This means that the .orElse() will return either a Result<string,string> or a Promise<Result<string,string>>
+   * Fortunately Promise.then() flattens nested Promises so that it all 'just works', but there should be a better
+   * way to do this.
+   *
+   * Also, how do you add type annotations to curried functions???
```
andrewrlee: Like this? https://github.com/jsdoc/jsdoc/issues/1286#issuecomment-341911947

server/services/caseListService.js
```diff
@@ -25,40 +33,74 @@ module.exports = function createCaseListService(nomisClientBuilder, roService, l
...
+   */
+  const getROCaseList = (username, token) => () =>
+    R.pipeWith(R.then, [
+      getStaffCodeFromDb,
+      staffCodeFromDb => staffCodeFromDb.orElse(() => getStaffCodeFromDelius(username)),
+      sc => sc.map(getOffendersForStaffCode(token)).match(R.identity, message => ({ hdcEligible: [], message })),
+    ])(username)
```
andrewrlee: I'd like to see the non-ramda version of this :-D

andrewrlee: would be tempted to extract out a method for getting staff code? maybe even push down to `roService` ? 
```javascript
 /** 
    @param {string} username
    @returns {Result<String, String>}
  */
const getStaffCode => {
  const staffCodeFromDb = await getStaffCodeFromDb(username)
  return staffCodeFromDb.isSuccess()
     ? staffCodeFromDb
     :  getStaffCodeFromDelius(username)
}

File: server/utils/Result.js [hunk 6]
```diff
@@ -0,0 +1,61 @@
+const Success = value => ({
```
andrewrlee: don't these need jsdoc parameters saying they are successFactories/FailFactories?

andrewrlee: or the the thing that the type exported is a ResultFactory?

server/services/caseListService.js
bell-pa: Good point!

test/services/caseListServiceTest.js
```diff
@@ -309,16 +316,44 @@ describe('caseListService', () => {
...
+        roService.getStaffByUsername = sinon.stub().rejects('Delius went bang!')
+
+        // Meh. Can't use chai's expect().to.throw() because this is async code.
+        try {
+          await service.getHdcCaseList(ROUser.token, ROUser.username, ROUser.role)
+        } catch (e) {
+          // The test passed!
+          return
+        }
+        expect.fail()
```
andrewrlee: there's some examples using chai: ```it('should throw if error in api when getting relationships if error status other than 404', () => {
      deliusClient.getAllOffenderManagers.rejects({ status: 401 })
      return expect(service.findResponsibleOfficer('123', 'token')).to.be.rejected()
    })```

PR #1038, opened 2019-12-12
Alee/dcs 280 types

andrewrlee:


server/utils/ResultTypes.d.ts
```diff
@@ -1,22 +1,22 @@
 export interface Result<S, E> {
-  value(): S
-  error(): E
+  success(): S
+  fail(): E
   map<T>(f: (v: S) => T): Result<T, E>
-  orElse(f: () => Result<S, E>): Result<S, E>
+  orElseTryAsync(f: () => Promise<this>): Promise<this>
```
andrewrlee: renamed to `orElseTryAsync`, `orElse` would generally have the type signature: `orElse(f: () => Result<S, E>): S`.

Using polymorphic `this` to return the current type

server/utils/ResultTypes.d.ts
```diff
@@ -1,22 +1,22 @@
 export interface Result<S, E> {
-  value(): S
-  error(): E
+  success(): S
```
andrewrlee: these differed between interface and impl

server/utils/Result.js
```diff
@@ -55,7 +65,12 @@ const Fail = error => ({
   },
 })
 
-module.exports = {
+/**
+ * @type {import("./ResultTypes").ResultFactory}
```
andrewrlee: this binds the type definitions to the types

server/utils/Result.js
```diff
@@ -8,11 +12,13 @@ const Success = value => ({
   },
 
   map(f) {
-    return Success(f(value))
+    const result = f(value)
+    // @ts-ignore
```
andrewrlee: this is the only `ts-ignore` I couldn't avoid - Success is of type `Success<S,E>`, we need to return `Success<T,E>` and I couldn't work out how to express that

server/services/caseListService.js
```diff
@@ -30,19 +35,18 @@ module.exports = function createCaseListService(nomisClientBuilder, roService, l
...
-   * staffCodeFromDB is a Result<string,string>, but getStaffCodeFromDelius() returns a Promise<Result<string,string>>
-   * This means that the .orElse() will return either a Result<string,string> or a Promise<Result<string,string>>
-   * Fortunately the 'await' prefix flattens nested Promises so that it all 'just works'.
-   *
    */
   const getROCaseList = (username, token) => async () => {
     const staffCodeFromDb = await getStaffCodeFromDb(username)
-    const staffCode = await staffCodeFromDb.orElse(() => getStaffCodeFromDelius(username))
+    const staffCode = await staffCodeFromDb.orElseTryAsync(() => getStaffCodeFromDelius(username))
     return staffCode.map(getOffendersForStaffCode(token)).match(R.identity, message => ({ hdcEligible: [], message }))
```
andrewrlee: so now intelliJ sees staffCode as type `Result<String, String>`.
Still need to add type info for the curried function `getOffendersForStaffCode`. 

This currently only compiles as `R.identity` moves everything over to `any`

server/services/caseListService.js
```diff
@@ -30,17 +41,12 @@ module.exports = function createCaseListService(nomisClientBuilder, roService, l
...
-   * Fortunately the 'await' prefix flattens nested Promises so that it all 'just works'.
-   *
    */
   const getROCaseList = (username, token) => async () => {
     const staffCodeFromDb = await getStaffCodeFromDb(username)
-    const staffCode = await staffCodeFromDb.orElse(() => getStaffCodeFromDelius(username))
-    return staffCode.map(getOffendersForStaffCode(token)).match(R.identity, message => ({ hdcEligible: [], message }))
+    const staffCode = await staffCodeFromDb.orElseTryAsync(() => getStaffCodeFromDelius(username))
+    const offendersForStaffCode = await staffCode.mapAsync(getOffendersForStaffCode(token))
+    return offendersForStaffCode.match(R.identity, message => ({ hdcEligible: [], message }))
```
andrewrlee: The async methods produce `Promise<Result<S,E>>` which will tend to break any method chaining. 

I guess there might be a way of keeping the `Promise` inside the `Result` but I suspect this would be much more  complicated.

PR #1040, opened 2019-12-13
DCS-262 Recording warnings when issues occur

andrewrlee:
This provide a general error/warnings queue to expose to the internal team.

This ticket covers storing warnings whenever a CA attempts to assign a case to an RO who has an unlinked account.
These ROs will not be notified that a case has been assigned to them - and some manual work needs to be done to fix the record.

There are other places where we merely log errors on things which end up biting us which could be surfaced in the queue and hopefully reduce the maintenance overhead of the service.

### DIFF
--- a/migrations/20191213164035_create-warning-table.js
+++ b/migrations/20191213164035_create-warning-table.js
@@ -1,0 +1,20 @@
+exports.up = knex =>
+  knex.schema.createTable('warnings', table => {
+    table.increments('id').primary('pk_warnings')
+    table
+      .timestamp('timestamp')
+      .notNullable()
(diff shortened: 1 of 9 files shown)


andrewrlee:
> LGTM.
> The unique constraint on booking_id + no DELETE statement made me think. It is possible for a CA to send a case to an RO more than once, but encountering errors both times has to be very unusual.

that's a good point. I would like to allow multiple types of unacknowledged warnings for the same booking to be allowed, but potentially duplicates of acknowledged ones. I've added a partial index to cover that.

Also replaced the switch with a simpler `if`!

PR #1041, opened 2019-12-16
DCS-262 Adding screens to allow displaying and acknowledgement of warnings

andrewrlee:


server/data/warningClient.js
```diff
@@ -17,14 +18,13 @@ const WarningClient = {
     return db.query(query)
   },
 
-  async acknowledgeWarning(errorId) {
+  async acknowledgeWarnings(errorIds) {
     const query = {
-      text: `UPDATE warnings SET acknowledged = true WHERE id=$1`,
-      values: [errorId],
+      text: format(`UPDATE warnings SET acknowledged = true WHERE id in (%L)`, errorIds),
```
brightonsbox: I guess we are in control of the errorIds so there's no injection attack vector here?

andrewrlee: pg-format is used to safely escape user input. We have to use it here as pg-node doesn't support array parameters (unless you use ANY)

PR #1043, opened 2019-12-18
DCS-264-feature flag created and tests updated

GurnankCheema:
This is the first part of the overall DCS-264 ticket.
Have created a continueCaToRoFeatureFlag key in config.js and passed the config object to the caService as an arg in the service building process in index.js.

The unit tests have been updated to include the the key. 
If the key is set to true, the caService will return 'null'
if the key is set to false, the caService will return the error codes or messages.

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -147,4 +147,6 @@ module.exports = {
     autostart: get('SCHEDULED_JOBS_AUTOSTART', 'no') === 'yes',
     overlapTimeout: get('SCHEDULED_JOBS_OVERLAP', 5000),
   },
+
+  continueCaToRoFeatureFlag: get('CONTINUE_CA_TO_RO_FEATURE_FLAG', 'yes'),
 }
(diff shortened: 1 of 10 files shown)


server/config.js
```diff
@@ -147,4 +147,6 @@ module.exports = {
     autostart: get('SCHEDULED_JOBS_AUTOSTART', 'no') === 'yes',
     overlapTimeout: get('SCHEDULED_JOBS_OVERLAP', 5000),
   },
+
+  continueCaToRoFeatureFlag: get('CONTINUE_CA_TO_RO_FEATURE_FLAG', 'yes'),
```
andrewrlee: generally nicer to coerce this into a boolean, like above: 
`autostart: get('SCHEDULED_JOBS_AUTOSTART', 'no') === 'yes'`

then you'd be able to:
```
if (!continueCaToRoFeatureFlag) {
        return null
 }
```

I think the logic is the wrong way round in your check? if the flag is enabled then you want to use the new feature.
So in dev you want to use the real check - whereas in production you'd want to always return null?

GurnankCheema: I have changed the code as per above but it doesn't seem logical to me that we should negate the continueCaToRoFeatureFlag in caService

PR #1045, opened 2019-12-18
DCS-286 Extend the Responsible Officer contact page to fetch details from Delius.

bell-pa:
Also:
* Bump the Docker file base image to node:10.17-slim from node:10.15-slim
* Remove the probation-teams service from docker-compose.yaml because it isn't used


server/routes/contact.js
```diff
@@ -1,21 +1,33 @@
...
+/**
+ *
+ * @param userAdminService
+ * @param {RoService} roService
+ * @returns {function(*): *}
+ */
+module.exports = (userAdminService, roService) => router => {
   router.get(
-    '/ro/',
+    '/:borkingId',
```
brightonsbox: borkingId?

andrewrlee: yeah.. maybe `associatedBookingId`?

types/probationTeams.d.ts
```diff
@@ -0,0 +1,3 @@
+export interface ProbationTeamsClient {
```
andrewrlee: this already exists in licences.d.ts

docker-compose.yml
```diff
@@ -30,29 +30,6 @@ services:
       - GLOBAL_SEARCH_URL=http://localhost:3002/global-search
       - ENABLE_TEST_UTILS=true
 
-  probation-teams:
```
andrewrlee: do we not want this any more?

types/probationTeams.d.ts
andrewrlee: oh no - I see you are just extracting it out.

Nothing wrong with that roContactDetailsService references the type from the probationteamclient directly - not sure if there is anything wrong with that either....

server/routes/contact.js
```diff
@@ -1,21 +1,33 @@
...
+      // because 'bookingId' is treated specially - see standardRouter.js
+      const { borkingId } = req.params
+      const [ro] = unwrapResult(await roService.findResponsibleOfficer(borkingId, res.locals.token))
 
-  router.get(
-    '/ro/:deliusUserId',
-    asyncMiddleware(async (req, res) => {
-      const { deliusUserId } = req.params
-      const contact = await userAdminService.getRoUserByDeliusId(deliusUserId)
+      const contact = ro.deliusId ? await userAdminService.getRoUserByDeliusId(ro.deliusId) : undefined
```
andrewrlee: meh.. is `ro.deliusId && await userAdminService.getRoUserByDeliusId(ro.deliusId)` equivalent?

server/routes/contact.js
```diff
@@ -1,21 +1,33 @@
...
-    '/ro/',
+    '/:borkingId',
     asyncMiddleware(async (req, res) => {
-      const roUsers = await userAdminService.getRoUsers()
-      return res.render('contact/roList', { roUsers })
-    })
-  )
+      // because 'bookingId' is treated specially - see standardRouter.js
+      const { borkingId } = req.params
+      const [ro] = unwrapResult(await roService.findResponsibleOfficer(borkingId, res.locals.token))
```
andrewrlee: `ro` would be undefined if an error is present?

server/routes/contact.js
bell-pa: You're right. I've added a test, fixed this - and an equivalent bug lower down.

server/routes/contact.js
bell-pa: yes. Fixed + test.

docker-compose.yml
bell-pa: discussed

PR #1046, opened 2019-12-19
DCS-292 Add RO's personal email to RO details (Delius)

bell-pa:


server/routes/contact.js
```diff
@@ -24,9 +24,15 @@ module.exports = (userAdminService, roService) => router => {
         return res.render('contact/ro', { contact })
       }
 
+      const functionalMailbox = ro && ro.lduCode && (await userAdminService.getFunctionalMailbox(ro.lduCode))
```
andrewrlee: This is very similar to what happens in `roContactDetailsService.getResponsibleOfficerWithContactDetails()`

might be worth seeing if that has all the data you need here?

PR #1047, opened 2019-12-21
DCS-267 Adding active LDU script

andrewrlee:
In order to prevent prison staff from processing cases for probation teams which aren't currently part of HDC, We need to populate a whitelist table with active LDU (local delivery unit) codes.

This script will return the LDUs for each licence that has been processed in the app. This will be the current LDU of the officer responsible for the offender - we will then compare these results with a list of LDUs for a probation area (officers could have moved teams or reassigned) to create a master list.

This PR also fixes an issue with the logger where `setLevels` on Winston lead to incorrect severity level reporting. This seems to be an issue with overwriting some monkey patching which the app insights instrumentation applies.  

(Also reducing the amount of logging which occurs when delius errors)

### DIFF
--- a/log.js
+++ b/log.js
@@ -1,17 +1,11 @@
-const winston = require('winston')
-const { AzureApplicationInsightsLogger } = require('winston-azure-application-insights')
+// eslint-disable-next-line import/order
 const appInsights = require('./azure-appinsights')
+const { AzureApplicationInsightsLogger } = require('winston-azure-application-insights')
+const winston = require('winston')
 const { flattenMeta } = require('./server/misc')
 
 const logger = new winston.Logger()
 
-logger.setLevels({
-  audit: 0,
-  error: 1,
-  warn: 2,
-  info: 3,
-  debug: 4,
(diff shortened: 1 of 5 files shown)


log.js
```diff
@@ -1,17 +1,11 @@
-const winston = require('winston')
-const { AzureApplicationInsightsLogger } = require('winston-azure-application-insights')
+// eslint-disable-next-line import/order
 const appInsights = require('./azure-appinsights')
+const { AzureApplicationInsightsLogger } = require('winston-azure-application-insights')
```
andrewrlee: ensuring app insights is registered first so it can do it's magical instrumentation on winston

server/data/audit.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
       throw new Error(`Unknown audit key: ${key}`)
     }
 
-    logger.audit('AUDIT', { key })
```
andrewrlee: this was only used in test mode

PR #1049, opened 2019-12-23
DCS-264: CA tasklist displays any error messages and also disables co…

GurnankCheema:
…ntinue button

http://localhost:3000/hdc/taskList/bookingId

The CA is prevented from progressing from from the tasklist if there are any errors during the eligibility stage.
This means the CA will not be able to complete the curfew address or handoff to the RO

The errors can result from the info in delius not being consistent or the  LDU not being active. 

The tasklist will display the error messages in a red border, similar to the GDS style but they  aren't hyperlinked because there is nothing on the page to link them to.

The Start now button is disabled if any errors are identified.

### DIFF
--- a/server/routes/config/taskListErrors.js
+++ b/server/routes/config/taskListErrors.js
@@ -1,0 +1,7 @@
+module.exports = {
+  NO_OFFENDER_NUMBER: 'Offender number required but it is not entered in Delius',
+  NO_COM_ASSIGNED: 'A Community Offender Manager has not been assigned. Please contact the Delius team',
+  LDU_INACTIVE: 'The Local Delivery Unit is in a geographical area that is currently inactive. Please refer via Nomis',
+  COM_NOT_ALLOCATED:
+    'The Community Offender Manager (COM) assigned is a pseudo. Please arrange for an actual COM to be assigned',
+}
--- a/server/routes/taskList.js
+++ b/server/routes/taskList.js
@@ -10,6 +10,8 @@ const { getAllowedTransition } = require('../utils/licenceStatusTransitions')
 const { isEmpty } = require('../utils/functionalHelpers')
 const getTaskListModel = require('./viewModels/taskListModels')
 const logger = require('../../log')
+const caTasks = require('../routes/viewModels/taskLists/caTasks')
+const taskListErrors = require('./config/taskListErrors')
 
 /**
  * @param {object} args
@@ -18,7 +20,8 @@ const logger = require('../../log')
  * @param {any} args.audit
  * @param {CaService} args.caService
  */
-module.exports = ({ prisonerService, licenceService, audit }) => router => {
+
+module.exports = ({ prisonerService, licenceService, audit, caService }) => router => {
   router.get(
     '/:bookingId',
     asyncMiddleware(async (req, res) => {
@@ -35,14 +38,15 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
       const licenceStatus = getLicenceStatus(licence)
       const allowedTransition = getAllowedTransition(licenceStatus, req.user.role)
       const { statusLabel } = getStatusLabel(licenceStatus, req.user.role)
+      let taskListModel
+      const errorCodes = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
+      const errorObject = errorCodes ? getAllErrorMessages(errorCodes) : null
 
-      const taskListModel = getTaskListModel(
-        req.user.role,
-        postRelease,
-        licenceStatus,
-        licence || {},
-        allowedTransition
-      )
+      if (licenceStatus.stage === 'ELIGIBILITY' && req.user.role === 'CA' && errorObject) {
+        taskListModel = caTasks.getCaTasksEligibilityLduInactive(licenceStatus)
+      } else {
+        taskListModel = getTaskListModel(req.user.role, postRelease, licenceStatus, licence || {}, allowedTransition)
+      }
 
       res.render('taskList/taskListBuilder', {
         licenceStatus,
@@ -54,6 +58,7 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
         bookingId,
         taskListModel,
         postApproval: ['DECIDED', 'MODIFIED', 'MODIFIED_APPROVAL'].includes(licenceStatus.stage),
+        errorObject,
       })
     })
   )
@@ -105,4 +110,12 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
   )
 
   return router
+}
+
+const getAllErrorMessages = errorCodes => {
+  const errorMsgs = Object.keys(errorCodes)
+  return {
+    errMsg1: taskListErrors[errorMsgs[0]],
+    errMsg2: taskListErrors[errorMsgs[1]],
+  }
 }
--- a/server/routes/taskList.js
+++ b/server/routes/taskList.js
@@ -10,7 +10,6 @@ const { getAllowedTransition } = require('../utils/licenceStatusTransitions')
 const { isEmpty } = require('../utils/functionalHelpers')
 const getTaskListModel = require('./viewModels/taskListModels')
 const logger = require('../../log')
-const caTasks = require('../routes/viewModels/taskLists/caTasks')
 const taskListErrors = require('./config/taskListErrors')
 
 /**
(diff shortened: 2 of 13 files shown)


server/routes/config/taskListErrors.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  NO_OFFENDER_NUMBER: 'Offender number required but it is not entered in Delius',
+  NO_COM_ASSIGNED: 'A Community Offender Manager has not been assigned. Please contact the Delius team',
+  LDU_INACTIVE: 'The Local Delivery Unit is in a geographical area that is currently inactive. Please refer via Nomis',
+  COM_NOT_ALLOCATED:
+    'The Community Offender Manager (COM) assigned is a pseudo. Please arrange for an actual COM to be assigned',
+}
```
andrewrlee: nice 👍 

Would be good for Louise to review or provide alternatives.. but obvs doesn't need to hold up this PR

server/routes/taskList.js
```diff
@@ -35,14 +38,15 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
       const licenceStatus = getLicenceStatus(licence)
       const allowedTransition = getAllowedTransition(licenceStatus, req.user.role)
       const { statusLabel } = getStatusLabel(licenceStatus, req.user.role)
+      let taskListModel
+      const errorCodes = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
```
andrewrlee: this should really only be called if `licenceStatus.stage === 'ELIGIBILITY' && req.user.role === 'CA'`.
It's quite a heavy weight call

server/services/caService.js
```diff
@@ -21,6 +21,7 @@ module.exports = function createCaService(roService, lduActiveClient, { continue
     async getReasonForNotContinuing(bookingId, token) {
       if (!continueCaToRoFeatureFlag) {
         return null
+        // continueCaToRoFeatureFlag is defaulted to false in config so !continueCaToRoFeatureFlag means CA should proceed to RO without checking for errors
```
andrewrlee: "defaulted false in config" is only true in certain envs and isn't static. 
If you wanted to describe this it would be nicer to raise the layer of abstraction: "When this feature is disabled, we never block the CA from continuing the case"

server/services/caService.js
```diff
@@ -30,9 +31,10 @@ module.exports = function createCaService(roService, lduActiveClient, { continue
           `Found reason for not continuing processing booking: ${bookingId}, error: ${error.code}:${error.message}`
         )
         switch (error.code) {
-          case NO_OFFENDER_NUMBER:
-          case NO_COM_ASSIGNED:
-            return error.code
+          case 'NO_OFFENDER_NUMBER':
+            return { NO_OFFENDER_NUMBER }
+          case 'NO_COM_ASSIGNED':
+            return { NO_COM_ASSIGNED }
```
andrewrlee: Not sure why this has changed? why return objects?

server/routes/taskList.js
```diff
@@ -106,3 +111,11 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
 
   return router
 }
+
+const getAllErrorMessages = errorCodes => {
+  const errorMsgs = Object.keys(errorCodes)
+  return {
+    errMsg1: taskListErrors[errorMsgs[0]],
+    errMsg2: taskListErrors[errorMsgs[1]],
```
andrewrlee: Think this would be nicer to represent as a list rather than an object - in the template you are only using list operators: isEmpty, iterate all values and it would be simpler to represent here too: 
`return errorCodes.map(error => taskListErrors[error])`

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -17,6 +17,35 @@ const createLicence = require('./tasks/createLicence')
 const finalChecks = require('./tasks/finalChecks')
 
 module.exports = {
+  getCaTasksEligibilityLduInactive: ({ optedOut, eligible, decisions, tasks }) => {
+    const { eligibility, optOut } = tasks
+    const eligibilityDone = eligibility === 'DONE'
+    const optOutUnstarted = optOut === 'UNSTARTED'
```
andrewrlee: there's no need for any of these booleans, we can see this as being a static view. 
We always want all these tasks to be visible.

server/routes/taskList.js
```diff
@@ -35,14 +38,15 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
       const licenceStatus = getLicenceStatus(licence)
       const allowedTransition = getAllowedTransition(licenceStatus, req.user.role)
       const { statusLabel } = getStatusLabel(licenceStatus, req.user.role)
+      let taskListModel
+      const errorCodes = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
+      const errorObject = errorCodes ? getAllErrorMessages(errorCodes) : null
```
andrewrlee: probably better to call this what it represents rather than the type = `errorMessages`

server/services/caService.js
```diff
@@ -41,12 +43,17 @@ module.exports = function createCaService(roService, lduActiveClient, { continue
...
+
       if (!isLduActive) {
-        return LDU_INACTIVE
+        return { LDU_INACTIVE }
       }
 
       if (!isAllocated) {
-        return COM_NOT_ALLOCATED
+        return { COM_NOT_ALLOCATED }
       }
```
andrewrlee: could potentially merge these 3 `if`s into:
```js
  return [
     ...(isLduActive ? [LDU_INACTIVE] : []),
     ...(isAllocated ? [COM_NOT_ALLOCATED] : [])
   ]
```

server/services/caService.js
GurnankCheema: ok

server/services/caService.js
GurnankCheema: ok

server/routes/taskList.js
GurnankCheema: ok

server/routes/taskList.js
GurnankCheema: ok

server/routes/taskList.js
GurnankCheema: ok

server/routes/config/taskListErrors.js
GurnankCheema: ok

server/services/caService.js
GurnankCheema: worked like this

 if (!isLduActive || !isAllocated) {
 return [...(!isLduActive ? [LDU_INACTIVE] : []), ...(!isAllocated ? [COM_NOT_ALLOCATED] : []) 
  ] }

server/routes/viewModels/taskLists/caTasks.js
GurnankCheema: I'm not so sure about this one. 
IWhen doing this section I followed the example of the getCaTasksEligibility function - line 49 - and it shows the visibility is not static because it too checks the eligibility and optOut

server/services/caService.js
andrewrlee: No - your approach of returning logical codes rather than content seems better.

server/routes/viewModels/taskLists/caTasks.js
andrewrlee: oh maybe we need to use that eligibility == done to prevent locking people out before they have finished the eligibility section? so rather than just checking is stage == eligible we need to check this as well (the bug we were talking about yesterday) - don't need to deal with this in this PR though.

Either way - the idea is that we only show this list of tasks when the user cannot do any additional tasks. So the list of task we display to the user should be static? we might be able to make this simpler than `getCaTasksEligiblity`. I might have got this wrong though.

server/routes/taskList.js
```diff
@@ -35,14 +38,19 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
...
-      const taskListModel = getTaskListModel(
-        req.user.role,
-        postRelease,
-        licenceStatus,
-        licence || {},
-        allowedTransition
-      )
+      if (licenceStatus.stage === 'ELIGIBILITY' && req.user.role === 'CA') {
+        const errorCodesList = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
+        errors = errorCodesList ? errorCodesList.map(error => taskListErrors[error]) : null
```
andrewrlee: think rather than default to null, this could default to an empty array - then you won't have to null check in the templates.

server/services/caService.js
andrewrlee: should be able to tidy this up by return empty array rather than null when we want to model "no messages" as `empty array` rather than `null`

server/views/taskList/taskListTemplate.pug
```diff
@@ -33,6 +33,15 @@ block content
   if !user.role
     p Error
   else
+    if errors
+    div.largeMarginBottom
+    if errors && errors.length !== 0
```
andrewrlee: once not relying on null to represent "no errors", we can remove this check

server/services/caService.js
GurnankCheema: done

server/services/caService.js
GurnankCheema: done

server/routes/viewModels/taskLists/caTasks.js
GurnankCheema: done as per ...
1. pass in any errors to to getTaskListModel() in server/routes/taskList.js
getTaskListModel() returns a fixed array if errors are passed in. The fixed array is similar to the one returned by  caTasks.getCaTasksEligibilityLduInactive() previously
2. removed the caTasks.getCaTasksEligibilityLduInactive() function completely as no longer needed

server/routes/taskList.js
```diff
@@ -35,13 +37,20 @@ module.exports = ({ prisonerService, licenceService, audit }) => router => {
       const licenceStatus = getLicenceStatus(licence)
       const allowedTransition = getAllowedTransition(licenceStatus, req.user.role)
       const { statusLabel } = getStatusLabel(licenceStatus, req.user.role)
+      let errors
+
+      if (licenceStatus.stage === 'ELIGIBILITY' && req.user.role === 'CA') {
+        const errorCodesList = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
+        errors = errorCodesList.length > 0 ? errorCodesList.map(error => taskListErrors[error]) : null
```
andrewrlee: The benefit of relying on empty array rather than null is that you don't need to handle the null here - just pass an empty array through. We're modelling no errors as empty array rather than null.

server/views/taskList/taskListTemplate.pug
andrewrlee: and by passing through empty array - we don't need to handle the null check here either.

server/routes/taskList.js
GurnankCheema: routes/taskList.js :- line 44, replaced ternary with if()
routes/viewModels/taskListModels.js (errors) changed to (errors && errors.length > 0)

server/views/taskList/taskListTemplate.pug
GurnankCheema: done

server/routes/taskList.js
GurnankCheema: done

PR #1050, opened 2019-12-24
DCS-293 Send emails for unlinked users

andrewrlee:
Allow sending email to clearing house, even if the responsible has an unlinked user and staff account

### DIFF
--- a/server/services/roContactDetailsService.js
+++ b/server/services/roContactDetailsService.js
@@ -41,6 +41,7 @@ module.exports = function createRoContactDetailsService(userAdminService, roServ
 
     return {
       ...deliusRo,
+      isUnlinkedAccount: false,
       email,
       functionalMailbox: orgEmail,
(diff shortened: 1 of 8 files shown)


server/services/roContactDetailsService.js
```diff
@@ -92,9 +95,10 @@ module.exports = function createRoContactDetailsService(userAdminService, roServ
 
       return {
         ...deliusRo,
+        isUnlinkedAccount: mailboxes.isUnlinkedAccount,
         email: mailboxes.email,
         functionalMailbox: mailboxes.functionalMailbox,
-        organisation: deliusRo.lduDescription, // TODO: check this
+        organisation: `${deliusRo.lduDescription} (${deliusRo.lduCode})`,
```
andrewrlee: changing this to match how we display org on ro contact details

PR #1051, opened 2019-12-30
Moving to jest

andrewrlee:
Moving away from mocha, chai and sinon and moving to jest instead.

Main benefits being better error reporting, more consistent with other dps projects.
It's also about 25% faster and has better support for typescript if we decide to go down that route.


server/data/dataAccess/db.js
```diff
@@ -11,7 +11,7 @@ const pool = new Pool({
   password: config.db.password,
   port: config.db.port,
   ssl:
-    config.db.sslEnabled === 'true'
+    config.production && config.db.sslEnabled === 'true'
```
andrewrlee: we now require db.js in test and mock out db.query in tests, so this actually fires. 
As we don't have a `.env` in CI and `sslEnabled` defaults to `true`, it will try and load a non existent file and fail all persistence tests.

andrewrlee:
> LGTM! 👍
> 
> Awesome work, don't really want to think about how long it took...

This tool helped alot! https://github.com/skovhus/jest-codemods
Then just a chunk of regex replacements

PR #1053, opened 2020-01-02
DCS-264: error length and margins

GurnankCheema:
1. Check the error array exists before gettings it's length property
2. Added a margin to the 'Check eligibility' row. This was present before but got removed along the way. Without the margin, the Start now button touches the horizontal rule

### DIFF
--- a/server/views/taskList/taskListTemplate.pug
+++ b/server/views/taskList/taskListTemplate.pug
@@ -33,7 +33,7 @@ block content
   if !user.role
     p Error
   else
-    if errors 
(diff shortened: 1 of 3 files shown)


server/views/taskList/taskListTemplate.pug
```diff
@@ -33,7 +33,7 @@ block content
   if !user.role
     p Error
   else
-    if errors 
+    if errors && errors.length > 0
```
andrewrlee: you don't need to check for errors as it will always be present - as we represent no errors as an empty array

PR #1054, opened 2020-01-03
DCS-265: new column in activeLdu table

GurnankCheema:
Inserted a new column in active_local_delivery_units table

The new column is probation_area_code.
This is needed to link an LDU to a probation area
Updated the addLdu and isLduPresent functions in the server/data/activeLduClient.js file as the save and search they perform need to include the probation area.

### DIFF
--- a/server/data/activeLduClient.js
+++ b/server/data/activeLduClient.js
@@ -1,19 +1,19 @@
 const db = require('./dataAccess/db')
 
 module.exports = {
-  async addLdu(lduCode) {
+  async addLdu(lduCode, probationAreaCode) {
     const query = {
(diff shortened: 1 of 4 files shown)


server/data/activeLduClient.js
```diff
@@ -1,19 +1,19 @@
 const db = require('./dataAccess/db')
 
 module.exports = {
-  async addLdu(lduCode) {
+  async addLdu(lduCode, probationAreaCode) {
```
andrewrlee: meh.. would generally order arguments in terms of size - as probation area is the more general, further arguments providing more context.

PR #1067, opened 2020-01-17
DCS-295 Adding delius hdc licences role

andrewrlee:
This role is added when a prison case administrator hands over a case to a responsible officer as we have identified that that delius user should have access to the licence service.

We add a delius role to the user. When that user attempts to access the service with their delius user details, nomis oauth server will map the new delius role to the standard licence responsible officer role.

The add role is an idempotent operation - so there is no problem executing it every time an RO is assigned a case.

### DIFF
--- a/test/data/deliusClient.test.js
+++ b/test/data/deliusClient.test.js
@@ -118,4 +118,18 @@ describe('deliusClient', () => {
       ])
     })
   })
+
+  describe('addResponsbileOfficerRole', () => {
+    test('should return data from api', () => {
+      fakeDelius.put(`/users/bobUser/roles/${config.delius.responsibleOfficerRoleId}`).reply(200, {})
+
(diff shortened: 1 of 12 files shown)


test/data/deliusClient.test.js
```diff
@@ -118,4 +118,18 @@ describe('deliusClient', () => {
       ])
     })
   })
+
+  describe('addResponsbileOfficerRole', () => {
+    test('should return data from api', () => {
+      fakeDelius.put(`/users/bobUser/roles/${config.delius.responsibleOfficerRoleId}`).reply(200, {})
+
+      return expect(deliusClient.addResponsibleOfficerRole('bobUser')).resolves.toStrictEqual({})
+    })
```
bell-pa: Why does the PUT return an empty object? Why return anything?

andrewrlee: now not returning anything from stubbing - response body still resolves to empty object

PR #1068, opened 2020-01-18
DCS 265 new UI for Probation and Ldu pages

GurnankCheema:
Two new pages added:

1. Probation areas: http://localhost:3000/admin/locations/probation-areas
This displays a list of hyperlinks for all the probation areas currently in Delius. Selecting a hyperlink takes the user to the next page that displays all the LDUs associated with that probation area. The url segment passes forward the code for the probation area . This list appears in alpha order.
The Probation area page is accessed via http://localhost:3000/admin/

2. Local Delivery Units page:
http://localhost:3000/admin/locations/probation-areas/Bir/local-delivery-units
Displays all the LDUs associated with a specific probation area. This page displays a checkbox for each LDU.
If there are no LDUs' to display, an information message is displayed.
The checkboxes will be preselected if that LDU is in the active_ldu table of the Licences DB.
Selecting or de-selecting a checkbox followed by save will update the DB.


server/services/lduService.js
```diff
@@ -13,19 +16,25 @@ const logger = require('../../log')
...
+      const probationAreas = await deliusClient.getAllProbationAreas()
+      sortObjArrayAsc(probationAreas)
+      return probationAreas
     },
 
     async getLdusForProbationArea(probationAreaCode) {
       const allLdus = await deliusClient.getAllLdusForProbationArea(probationAreaCode)
       const activeLdus = await activeLduClient.allActiveLdusInArea(probationAreaCode)
       const activeLdusCodes = activeLdus.map(ldu => ldu.code)
+      const probationAreaDescription = getIn(allLdus, [0, 'description'])
```
bell-pa: getAllLdusForProbationArea returns an array of Ldu. Each Ldu has code and description of that LDU. I don't understand how you can get a probationAreaDescription from the first Ldu.
Also, below, since an Ldu does not have a 'teams' field I don't see how you can extract a team from the first element of the allLdus array.  I suspect you've mixed up the 'teams' within one LDU with the list of LDUs.

Probably want to do something like 
allLdus.map(ldu => ({
  code: ldu.code,
  description: ldu.description,
  active: activeLduCodes.includes(ldu.code)
}))

GurnankCheema: The 'teams' came from the obj Delius is returning.... I think ...
I got it from where they had this example

[   {     "code": "N01",     "description": "NPS North West",     "teams": [       {   "code": "string",         "description": "string",    plus more stuff  }]   

plus more stuff
 } ]
--

server/routes/admin/locations.js
```diff
@@ -13,26 +13,26 @@ module.exports = lduService => (router, audited) => {
...
+      res.render('admin/locations/ldus', { ldus, msg: req.flash('success') })
     })
   )
 
   router.post(
     '/probation-areas/:probationAreaCode/local-delivery-units',
     audited,
     asyncMiddleware(async (req, res) => {
-      const { activeLdus } = req.body
+      const activeLdus = req.body.activeLdus || []
```
andrewrlee: meh.. this is fine, to have got it working with destructuring you could have:
`const { activeLdus = [] } = req.body`

server/services/lduService.js
```diff
@@ -13,19 +14,27 @@ const logger = require('../../log')
 module.exports = function createLduService(deliusClient, activeLduClient) {
   return {
     async getAllProbationAreas() {
-      return deliusClient.getAllProbationAreas()
+      const { content: probationAreas } = await deliusClient.getAllProbationAreas()
+      probationAreas.sort((a, b) => a.description.localeCompare(b.description, 'en', { ignorePunctuation: true }))
+      return probationAreas
     },
 
     async getLdusForProbationArea(probationAreaCode) {
```
andrewrlee: The return type doesn't make sense based on the methods names.

A bit of modelling here would be good. 

We could say that the object we you are returning represents a probation area.
A probation area consists of a list of a description and a list of LDUs

```js
{
  description: 'some description',
  ldus: [
    ...  some ldus 
 ]
}
```

server/services/lduService.js
GurnankCheema: done

server/routes/admin/locations.js
GurnankCheema: done

PR #1072, opened 2020-01-23
DCS-328: Ldus reorganised and  duplicates removed

GurnankCheema:
http://localhost:3000/admin/locations/probation-areas
Probation area code appended to name

http://localhost:3000/admin/locations/probation-areas/:code/local-delivery-units
Ldus filtered by code to remove any duplicates. Displayed only first description where the same ldu may have more than one description

Extra padding to bottom of save button

### DIFF
--- a/server/services/lduService.js
+++ b/server/services/lduService.js
@@ -29,23 +29,21 @@ module.exports = function createLduService(deliusClient, activeLduClient) {
 
       const { content: ldus = [] } = await deliusClient.getAllLdusForProbationArea(probationAreaCode)
 
-      const allLdus = ldus
-        .map(ldu => ({
-          code: ldu.code,
-          description: ldu.description,
-          active: activeLdusCodes.includes(ldu.code),
-        }))
-        .sort((a, b) => a.description.localeCompare(b.description, 'en', { ignorePunctuation: true }))
+      const allLdus = ldus.map(ldu => ({
+        code: ldu.code,
+        description: ldu.description,
+        active: activeLdusCodes.includes(ldu.code),
+      }))
 
-      const uniqueLdus = Object.entries(R.groupBy(ldu => ldu.code, allLdus)).map(ldu => ldu[1][0])
+      const uniqueLdus = [...Object.entries(R.groupBy(R.prop('code'), allLdus))]
+        .sort(([lduCode1], [lduCode2]) => lduCode1.localeCompare(lduCode2))
+        .map(([_, ldu]) => ldu.sort((ldu1, ldu2) => ldu1.description.localeCompare(ldu2.description))[0])
 
-      const probationArea = {
+      return {
         code: probationAreaCode,
         description: probationAreaDescription,
         ldus: uniqueLdus,
       }
-
-      return probationArea
     },
 
     async updateActiveLdus(probationAreaCode, activeLdus) {
--- a/server/services/lduService.js
+++ b/server/services/lduService.js
@@ -1,3 +1,4 @@
+const R = require('ramda')
 const logger = require('../../log')
 
 /**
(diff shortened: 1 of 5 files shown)


server/services/lduService.js
```diff
@@ -28,16 +29,20 @@ module.exports = function createLduService(deliusClient, activeLduClient) {
...
-      }))
+      const allLdus = ldus
+        .map(ldu => ({
+          code: ldu.code,
+          description: ldu.description,
+          active: activeLdusCodes.includes(ldu.code),
+        }))
+        .sort((a, b) => a.description.localeCompare(b.description, 'en', { ignorePunctuation: true }))
+
+      const uniqueLdus = Object.entries(R.groupBy(ldu => ldu.code, allLdus)).map(ldu => ldu[1][0])
```
bell-pa: There's no ordering on Object.entries so your previous sort is lost.
Also the ldu's are in an arbitrary order within each group.  They need to be sorted by description before selecting the first.

Try something like this:
```javascript
const lduMap = R.groupBy(R.prop('code'), ldus)
Object
  .entries(lduMap)
  .sort((a, b) => a[0].localeCompare(b[0]))
  .map(entry => entry[1].sort((a,b) => a.description.localeCompare(b.description))[0])
```
Here's an [example of how that works](https://ramdajs.com/repl/#?const%20ldus%20%3D%20%5B%0A%20%20%7B%20code%3A%20%27A%27%2C%20description%3A%20%27An%20A%27%7D%2C%0A%20%20%7B%20code%3A%20%27C%27%2C%20description%3A%20%27A%20c%27%20%7D%2C%0A%20%20%7B%20code%3A%20%27B%27%2C%20description%3A%20%27A%20B%27%20%7D%2C%0A%20%20%7B%20code%3A%20%27A%27%2C%20description%3A%20%27A%20a%27%20%7D%2C%0A%20%20%7B%20code%3A%20%27A%27%2C%20description%3A%20%27An%20a%27%7D%2C%0A%20%20%7B%20code%3A%20%27B%27%2C%20description%3A%20%27A%20b%27%20%7D%2C%0A%20%20%7B%20code%3A%20%27A%27%2C%20description%3A%20%27A%20A%27%20%7D%2C%0A%5D%0A%0Aconst%20lduMap%20%3D%20R.groupBy%28R.prop%28%27code%27%29%2C%20ldus%29%0AObject%0A%20%20.entries%28lduMap%29%0A%20%20.sort%28%28a%2C%20b%29%20%3D%3E%20a%5B0%5D.localeCompare%28b%5B0%5D%29%29%0A%20%20.map%28entry%20%3D%3E%20entry%5B1%5D.sort%28%28a%2Cb%29%20%3D%3E%20a.description.localeCompare%28b.description%29%29%5B0%5D%29%0A) That you can experiment with.

I'd suggest extracting the code which transforms `allLdus' into `uniqueLdus` into a function, then writing some tests for that function to show that it does what you expect of it.

GurnankCheema: done

test/services/lduService.test.js
```diff
@@ -59,8 +59,76 @@ describe('lduService', () => {
...
+          { code: 'wtl', description: 'Waterloo2' },
+        ],
+      })
+
+      deliusClient.getAllProbationAreas.mockResolvedValue({ content: [{ code: 'Lon', description: 'London' }] })
+      activeLduClient.allActiveLdusInArea.mockResolvedValue([{ code: 'ham' }, { code: 'pic' }])
+
+      const result = await lduService.getProbationArea('Lon')
+
+      expect(result).not.toEqual({
```
andrewrlee: This isn't a very persuasive test. There's literally an infinite amount of things that results is not equal to. It's not equal to: `"some random string"`, `1`, `true`, a literal fish, the Milk Way, the concept of silence... etc..

Instead it would be more persuasive to show what it is equal to and then the reader can see that there are no duplicates in there, for themselves.

andrewrlee: which it looks like you already have - so this test is unnecessary?

server/services/lduService.js
```diff
@@ -34,13 +35,15 @@ module.exports = function createLduService(deliusClient, activeLduClient) {
         active: activeLdusCodes.includes(ldu.code),
       }))
 
-      const probationArea = {
+      const uniqueLdus = Object.entries(R.groupBy(R.prop('code'), allLdus))
+        .sort((a, b) => a[0].localeCompare(b[0]))
+        .map(entry => entry[1].sort((a, b) => a.description.localeCompare(b.description))[0])
```
andrewrlee: destructuring and using appropriate variable names to represent what you are comparing might make this a bit clearer: 

```js
 .sort(([lduCode1], [lduCode2]) => lduCode1.localeCompare(lduCode2))
.map([code, ldus]) => ldus.sort((ldu1, ldu2) => ldu1.description.localeCompare(ldu2.description))[0])
```

test/services/lduService.test.js
GurnankCheema: I did this test  to show that the positive test wasn't passing for all cases.
Now removed

server/services/lduService.js
GurnankCheema: I'm not sure it makes things that much clearer although I think the it's okay to destructure. 

However passing in an array to map([code, ldus]) causes problems with the linter because we don't actually use 'code' , the first item in the array. It's there just because we want to access the second item (descriptions of the ldu), I think ??

But this would work: 
const uniqueLdus = [...Object.entries(R.groupBy(R.prop('code'), allLdus))]
  .sort(([lduCode1], [lduCode2]) => lduCode1.localeCompare(lduCode2))
  .map( ldu => ldu[1].sort((ldu1, ldu2) => ldu1.description.localeCompare(ldu2.description))[0])

GurnankCheema: done

PR #1079, opened 2020-01-28
DCS 318: new licence additional condition added to the drugs section

GurnankCheema:
A new condition asking about drugs dependency added to the Drugs, Health and Behaviour section.
The new content is the same as that at the bottom of the page under the Post-sentence section.
The printed licences display the new content either in the main section or the PSS section, depending on the licence type


server/views/licenceConditions/includes/appointmentDetailsInDrugsSection.pug
```diff
@@ -0,0 +1,24 @@
+div.pure-g
+  div.pure-u-1
+    //label(for='appointmentDateInDrugsSection') Date
+    //input(type='text' name='appointmentDateInDrugsSection' id='appointmentDateInDrugsSection' class='form-control' value=condition.user_submission && condition.user_submission.appointmentDateInDrugsSection aria-label="Appointment date")
```
andrewrlee: can we get rid of this commented code

andrewrlee: (In a follow up is fine)

GurnankCheema: will do

PR #1080, opened 2020-01-28
DCS-331 backend work to support admin staff sending emails to ROs

andrewrlee:
This will allow a frontend to be developed to allow sending emails to new linked RO users.
The original work to transition the state and reset any decisions will have already occurred when the CA handed over the case.

The original handover will have triggered a SEND audit event using the PCA username.
This will trigger a NOTIFY event with the admin staffs username.
It will also attempt to assign the HDC Update role to the RO's delius user.

This new service method will return the RO to provide some contextual information on success.
Otherwise it will return the error for display to the user.
Unlike the existing service method it will error if the user is not linked - the assumption will
be that this will only be called once the user is linked


server/views/admin/licences/activity.pug
```diff
@@ -19,10 +19,18 @@ mixin send(details)
...
+          | Sent from RO → CA (Opted out)
       when 'caToDm'
           | Sent from CA → DM
+      when 'caToDmRefusal'
+          | Sent from CA → DM (Refusal)
       when 'dmToCa'
           | Sent from DM → CA
+      when 'dmToCaReturn'
+          | Sent from DM → CA (Return)
       default
```
andrewrlee: These weren't mapped so events with these transitions appeared blank in the admin audit history screen

PR #1087, opened 2020-02-04
DCS-351 Relax restriction on contact page,

andrewrlee:
Previously the accessing user required access to the caselist associated with the booking to view the responsible RO. 

The only info we use from NOMIS in this call is to map from booking id to nomis id, no personal info.

### DIFF
--- a/server/authentication/authInit.js
+++ b/server/authentication/authInit.js
@@ -27,7 +27,7 @@ module.exports = (userService, audit) => {
       const user = await getUser(accessToken, refreshToken, params.expires_in, params.user_name)
       return done(null, user)
     } catch (error) {
-      logger.error('Sign in error ', error.stack)
+      logger.error(`Sign in error for user: '${params.user_name}'`, error.stack)
       return done(null, false, { message: 'A system error occurred; please try again later' })
     }
   }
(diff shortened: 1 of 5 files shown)


server/authentication/authInit.js
```diff
@@ -27,7 +27,7 @@ module.exports = (userService, audit) => {
       const user = await getUser(accessToken, refreshToken, params.expires_in, params.user_name)
       return done(null, user)
     } catch (error) {
-      logger.error('Sign in error ', error.stack)
+      logger.error(`Sign in error for user: '${params.user_name}'`, error.stack)
```
andrewrlee: Adding user name to log out put to allow detecting which user tries to access licences without having the correct roles.

test/supertestSetup.js
```diff
@@ -217,6 +217,14 @@ const setup = {
     app.use(bodyParser.json())
     app.use(bodyParser.urlencoded({ extended: false }))
     app.use(prefix, route)
+    app.use((error, req, res, next) => {
+      if (error.status !== 403) {
+        // eslint-disable-next-line no-console
+        console.log('an error occurred:', error)
```
andrewrlee: Really annoying trying to find out why integration tests fail without this!

PR #1089, opened 2020-02-05
DCS-347: user must select a radio button before they can move on

GurnankCheema:
http://localhost:3000/hdc/proposedAddress/rejected/{bookingId}

Previously user could 'save and continue' without selecting the yes/no radio. This was incorrect  behaviour so have added validation checking. Now user is redirected back to the same page until they make a selection

### DIFF
--- a/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
+++ b/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
@@ -1,0 +1,77 @@
+package uk.gov.justice.digital.hmpps.licences.specs.assessment
+
+import geb.spock.GebReportingSpec
+import spock.lang.Shared
+import spock.lang.Stepwise
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.ProposedAddressRejectedPage
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.ProposedAddressCurfewAddressPage
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.ProposedAddressCurfewAddressChoicePage
+
+import uk.gov.justice.digital.hmpps.licences.util.Actions
+import uk.gov.justice.digital.hmpps.licences.util.TestData
+
+@Stepwise
+class ProposedAddressRejectedSpec extends GebReportingSpec {
+
+    @Shared
+    TestData testData = new TestData()
+
+    @Shared
+    Actions actions = new Actions()
+
+    def setupSpec() {
+      testData.loadLicence('assessment/unstarted')
+      actions.logIn('CA')
+    }
+
+    def cleanupSpec() {
+      actions.logOut()
+    }
+
+    def "Click 'save and continue' without selecting a radio button"() {
+
+        given: 'I am on the ProposedAddressRejectedPage'
+        to ProposedAddressRejectedPage, testData.markAndrewsBookingId
+
+        when: 'I select the continue button'
+        find('#continueBtn').click()
+
+        then: 'I am redirected to the same page and error messages are displayed'
+        at ProposedAddressRejectedPage
+
+        errorSummary.text().contains('There is a problem')
+        errorSummaryLinkMessage.text().contains('Select yes or no')
+        inlineErrorMessage.text().contains("Select yes or no")
+    }
+
+    def "Select the 'Yes' radio then 'save and continue'"() {
+
+        given: 'I am on the ProposedAddressRejectedPage'
+        to (ProposedAddressRejectedPage, testData.markAndrewsBookingId)
+
+        when: "I select the Yes radio button"
+        alternativeAddressRadios.value('Yes')
+
+        and : 'I select the continue button'
+        find('#continueBtn').click()
+
+        then: 'I am taken to the curfew address page'
+        at ProposedAddressCurfewAddressPage
+    }
+
+    def "Select the 'No' radio then 'save and continue'"() {
+
+        given: 'I am on the ProposedAddressRejectedPage'
+        to ProposedAddressRejectedPage, testData.markAndrewsBookingId
+
+        when: "I select the Yes radio button"
+        alternativeAddressRadios.value('No')
+
+        and : 'I select the continue button'
+        find('#continueBtn').click()
+
+        then: 'I am taken to the curfew address page'
+        at ProposedAddressCurfewAddressChoicePage
+    }
+}
+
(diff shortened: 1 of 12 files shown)


licences-specs/src/test/resources/GebConfig.groovy
```diff
@@ -26,8 +26,8 @@ driver = {
   // System.setProperty('webdriver.chrome.driver', '/Applications/chromedriver')
   // System.setProperty('webdriver.chrome.driver', '/usr/local/bin/chromedriver')
   ChromeOptions options = new ChromeOptions()
-  options.addArguments('headless')
-//  options.addArguments('chrome')
+ options.addArguments('headless')
+  // options.addArguments('chrome')
```
andrewrlee: guess this file should just be reverted?

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+      actions.logOut()
+    }
+
+    def "Click 'save and continue' without selecting a radio button"() {
+
+        given: 'I am on the ProposedAddressRejectedPage'
+        to ProposedAddressRejectedPage, testData.markAndrewsBookingId
+
+        when: 'I select the continue button'
+        find('#continueBtn').click()
```
andrewrlee: These test look good.

Is there any reason why this button can't be on the ProposedAddressRejectedButton?

Generally you should be interacting via the page object rather than doing bespoke calls to target elements directly.

server/routes/address.js
```diff
@@ -57,6 +57,17 @@ module.exports = ({ licenceService, nomisPushService }) => (router, audited, { p
       const { enterAlternative, bookingId } = req.body
       const { licence } = res.locals.licence
 
+      const errorObject =
+        licenceService.validateForm({
+          formResponse: req.body,
+          pageConfig: formConfig.rejected,
+        }) || {}
+
+      if (errorObject.enterAlternative) {
```
andrewrlee: Not sure if you are following a pattern but is there a reason to be specific about the type of error?  

Would it be valid to just check if there are *any* errors here?

server/views/proposedAddress/rejected.pug
```diff
@@ -7,12 +7,21 @@ block content
...
     h2.heading-large Curfew address rejected
     p The following address has been returned by the Responsible Officer:
+      if errorMessage
+        div.pure-g
+          div.error-summary.pure-u-1.pure-u-md-1-2(role="alert" aria-labelledby="error-summary-heading" tabindex="-1")
+            h2.heading-medium.error-summary-heading#error-summary-heading
+              | There is a problem
+            ul.error-summary-list
+                li
+                  a(href="#enterAlternativeYes")  #{errorMessage}
```
andrewrlee: there are (several for some reason?!) existing macros for managing errors - would it be possible to use one of those?

server/routes/address.js
```diff
@@ -57,6 +57,17 @@ module.exports = ({ licenceService, nomisPushService }) => (router, audited, { p
       const { enterAlternative, bookingId } = req.body
       const { licence } = res.locals.licence
 
+      const errorObject =
+        licenceService.validateForm({
```
andrewrlee: can we add a route test for this too?

licences-specs/src/test/resources/GebConfig.groovy
GurnankCheema: done

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
GurnankCheema: agreed. Done

server/routes/address.js
GurnankCheema: I did start off by just checking for the errorObject but always got an entry for csrf and bookingId included in the errorObject, like this:
{ 
   enterAlternative: 'Select yes or no',
  _csrf: '"_csrf" is not allowed',
  bookingId: '"bookingId" is not allowed' 
}

So, I tried removing crsf and bookingId from server/views/includes/alternativeForm.pug but that caused the app to route to http://localhost:8080/auth/login

Hence I thought I should only do a conditional check for errorObject.enterAlternative

So, going back to your suggestion, I will not specifically check for the enterAlternative in the error Object but rather filter out the csrf and bookingId.

server/views/proposedAddress/rejected.pug
GurnankCheema: Regarding the inline error message in alternativeForm.pug
There wasn't a mixin that applied the form-group-error class which paints a red vertical line on LHS of inputs. In fact in other modules, eg server/views/approval/release.pug,  the class is applied in a similar way to that used in the code above

Regarding error summary box in rejected.pug
The nearest fit was the server/views/includes/errorBannerWithDetailObject.pug mixin

server/routes/address.js
GurnankCheema: I have assumed you mean a GEB test to check that the user lis taken from http://localhost:3000/hdc/taskList/{bookingId}  to server/routes/address.js if the RO has rejected the address??

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
andrewrlee: Seems like you've updated other places just not this one?

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+    def "Select the 'Yes' radio then 'save and continue'"() {
+
+        given: 'I am on the ProposedAddressRejectedPage'
+        to (ProposedAddressRejectedPage, testData.markAndrewsBookingId)
+
+        when: "I select the Yes radio button"
+        alternativeAddressRadios.value('Yes')
+
+        and : 'I select the continue button'
+        proposedAddressRejectedButton.click()
```
andrewrlee: `continueButton` might be a better name for this variable? (see the line above)

server/routes/address.js
andrewrlee: Just a normal Jest route test, e.g: risk.test.js or any of the ones in here: 
`test/routes`.

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
GurnankCheema: oops! 
done

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/ProposedAddressRejectedSpec.groovy
GurnankCheema: done

server/routes/address.js
GurnankCheema: done

PR #1095, opened 2020-02-14
DCS-358 Raising app insights event when case is handed over.

andrewrlee:
This event will contain information about the source and target of the case.
This should allow us to easily see frequency at which certain prison/LDUs handover cases - see which prisons are using the service and which ones aren't.

Also adding this information to the audit trail in case we want to surface this through the
app at later date.

Whilst doing this I noticed that we weren't capturing any audit info for DMs - we're now doing that.


server/services/notifications/notificationService.js
```diff
@@ -24,15 +25,27 @@ module.exports = function createNotificationService(
   }
 
   async function sendCaOrDm({ transition, bookingId, token, licence, prisoner, user }) {
-    const submissionTarget = await prisonerService.getOrganisationContactDetails(transition.receiver, bookingId, token)
+    const { submissionTarget, source, target } = await prisonerService.getDestinations(
```
bell-pa: const destinations = ?

server/services/notifications/notificationService.js
```diff
@@ -24,15 +25,27 @@ module.exports = function createNotificationService(
...
     if (transition.type === 'dmToCaReturn') {
       await licenceService.removeDecision(bookingId, licence)
     }
 
-    auditEvent(user.username, bookingId, transition.type, submissionTarget)
+    await eventPublisher.raiseCaseHandover({
+      username: user.username,
+      bookingId,
+      transitionType: transition.type,
+      submissionTarget,
```
bell-pa: ...destinations ?

server/index.js
```diff
@@ -75,22 +77,24 @@ const caAndDmNotificationSender = createCaAndDmNotificationSender(
   config
 )
 
+const eventPublisher = new EventPublisher(audit, appInsights)
```
bell-pa: Good idea.

andrewrlee: Yes, I could have done this - was just being a bit more specific. I'll have a look in a follow up

PR #1097, opened 2020-02-17
DCS-360: remove warning messages in DM tasklist

GurnankCheema:
http://localhost:3000/hdc/taskList/bookingId
http://localhost:3000/hdc/approval/release/bookingId

If a DM rejects a proposed address they hand-off back to the CA. The CA can then propose a new address and refer it back to DM again (passing through RO first) for re-consideration.
When the DM logs in, they see a warning message that relates to their previous decision and because the CA has proposed a new address, the warning message should not be displayed.
Code changed so that the DM does not see the warning message or the inline message in the final task in the tasklist.
The DMs previous rejection reasons also removed because the DM needs to re-adjudicate.

Changes as follows:
1.  server/routes/approval.js to convert the previous decision to {}
2.  server/routes/viewModels/taskLists/dmTasks.js . to remove the second part of the label applied to the final task in the DM's tasklist
3.  server/views/taskList/taskListTemplate.pug to remove the warning message from the DM tasklist.
Note - the warning message is still required by the CA tasklist
4. licences/specs/decision/ApprovalSpec.groovy GEB test updated


GurnankCheema:
new PR created using new branch

PR #1107, opened 2020-02-21
DCS-381 Update Ineligibility and Presumed Unsuitability Reasons

bell-pa:


server/routes/eligibility.js
```diff
@@ -2,7 +2,7 @@ const formConfig = require('./config/eligibility')
 const { asyncMiddleware } = require('../utils/middleware')
 const { getIn, firstItem } = require('../utils/functionalHelpers')
 const createStandardRoutes = require('./routeWorkers/standard')
-const { cja2003Sza } = require('../config')
+const { cja2003s19za } = require('../config')
```
andrewrlee: and I thought this was catchy before! :-D

server/routes/eligibility.js
```diff
@@ -58,9 +49,7 @@ module.exports = ({ licenceService, nomisPushService }) => (router, audited, con
     asyncMiddleware(standard.callbackPost('suitability', checksPassedCallback))
   )
 
-  router.get('/excluded/:bookingId', asyncMiddleware(get))
```
andrewrlee: just checking this is intentional
So you don't need to treat excluded as a special case - you are treating all eligibilty routes in the same way?

bell-pa: Yes. I'm using a copy of standard.js 'get'. Only change is adding the cja2003Sza flag for the views.  The 'suitability' form also needs this flag now because Louise added another option to this form: 'Offender not currently serving for a terrorism or terrorism-related offence but with a history of such offending'

PR #1109, opened 2020-02-24
DCS-349 Display stringified query string

andrewrlee:


server/services/config/formConfig.js
```diff
@@ -1,5 +1,3 @@
-const config = require('../../config')
```
andrewrlee: unused var

PR #1111, opened 2020-02-27
DCS-384 Fixing issue with bad request when getting sentences from elite

andrewrlee:
If we don't specify offender numbers when requesting sentences, elite throws a 400 bad request.
We had a guard to protect against this but it assumed that offender numbers would always be
present in delius for each prisoner and this is not always the case.

Offenders without offender numbers would be filtered out of the request but
If all prisoners did not have offender numbers then this would result in a 400

This fixes the guard to prevent this edgecase.

### DIFF
--- a/server/services/roService.js
+++ b/server/services/roService.js
@@ -61,8 +61,8 @@ module.exports = function createRoService(deliusClient, nomisClientBuilder) {
     async getROPrisoners(staffCode, token) {
       const nomisClient = nomisClientBuilder(token)
       const requiredPrisoners = await getROPrisonersFromDelius(staffCode)
-      if (!isEmpty(requiredPrisoners)) {
-        const requiredIDs = requiredPrisoners.map(prisoner => prisoner.nomsNumber)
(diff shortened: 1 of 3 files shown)


server/services/roService.js
```diff
@@ -61,8 +61,8 @@ module.exports = function createRoService(deliusClient, nomisClientBuilder) {
     async getROPrisoners(staffCode, token) {
       const nomisClient = nomisClientBuilder(token)
       const requiredPrisoners = await getROPrisonersFromDelius(staffCode)
-      if (!isEmpty(requiredPrisoners)) {
-        const requiredIDs = requiredPrisoners.map(prisoner => prisoner.nomsNumber)
+      const requiredIDs = requiredPrisoners.filter(prisoner => prisoner.nomsNumber).map(prisoner => prisoner.nomsNumber)
+      if (!isEmpty(requiredIDs)) {
```
bell-pa: Or push this check into nomisClient.getOffenderSentencesByNomisId ()
if 0 nomsNumbers then short circuit and return [] ?

andrewrlee: Yes, will do as part of next ticket which is in the same area

PR #1113, opened 2020-02-28
DCS-385 Adding batching of requests to when retrieving large numbers of sentences

andrewrlee:
NOMIS returns a 400 "Header too large" response when ROs are assigned a large list of offenders. This is due to coercing a large list of nomis ids into a very long query string.

ROs generally shouldn't be assigned large lists of offenders but it seems a small number of LDUs have strange ways of working and do end up in this state.

We could have potentially added a POST endpoint to handle lots of offender Nos but this seemed a bit overkill for such a small edge case.

### DIFF
--- a/server/data/nomisClientBuilder.js
+++ b/server/data/nomisClientBuilder.js
@@ -11,7 +11,7 @@ const moment = require('moment')
 const superagent = require('superagent')
 const logger = require('../../log')
 const config = require('../config')
-const { merge, pipe, getIn } = require('../utils/functionalHelpers')
+const { merge, pipe, getIn, splitEvery, isEmpty } = require('../utils/functionalHelpers')
 const { unauthorisedError } = require('../utils/errors')
 
 const timeoutSpec = {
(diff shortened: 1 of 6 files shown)


server/data/nomisClientBuilder.js
```diff
@@ -80,12 +91,17 @@ module.exports = token => {
...
-      const query = { offenderNo: nomisIds }
-      const headers = { 'Page-Limit': 10000 }
+      if (isEmpty(nomisIds)) {
+        return []
+      }
+
+      const prisoners = await batchRequests(nomisIds, batchSize, batch => {
+        const query = { offenderNo: batch }
+        return nomisGet({ path, query, headers: { 'Page-Limit': batchSize } })
+      })
```
timharrison-moj: I just did something similar for pathfinder, and for the same endpoint! It seemed pretty happy with much larger batch sizes than the default 50 here. In PF it was more than happy at 500 - but just informational - no problem with batch sizes lower than this.

PR #1126, opened 2020-03-20
DCS-224 Fix first night curfew hours.

bell-pa:
Was surprisingly simple in the end.
More or less reverted a change that Yasin made a long time ago, but classified the first-night config data  in pdfData.js as `mandatoryPreRelease` and tweaked a bit of filtering in pdfFormatter.js
Using the `mandatoryPreRelease` value means that the first-night values can be excluded from the list of mandatory values for 'vary' cases (post release), which in turn means that when those values are missing they don't prevent creating a PDF licence.
Also tweaked some code in the process of trying to work out how it works.  No idea if its any easier to comprehend, but I understand it better now.

### DIFF
--- a/server/services/utils/pdfFormatter/pdfFormatter.js
+++ b/server/services/utils/pdfFormatter/pdfFormatter.js
@@ -1,12 +1,15 @@
+const R = require('ramda')
 const moment = require('moment')
 const romanise = require('romannumerals')
-const { getIn, isEmpty, mergeWithRight } = require('../../../utils/functionalHelpers')
+const { isEmpty, mergeWithRight } = require('../../../utils/functionalHelpers')
 const pdfData = require('../../config/pdfData')
 const config = require('../../../config')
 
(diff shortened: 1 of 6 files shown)


server/services/utils/pdfFormatter/pdfFormatter.js
```diff
@@ -32,76 +36,73 @@ function formatPdfData(
...
+            missing: {
+              [spec.group]: {
+                [spec.required]: {
+                  [key]: spec.displayName,
+                },
+              },
+            },
+          }
+
+      return mergeWithRight(summary, newObject)
```
andrewrlee: is `mergeWithRight` equivalent to : `{...summary, ...newObject }` ?

bell-pa: No. Its a deep merge.  Spreading the objects overwrites nested data whereas mergeWithRight adds to existing data.

PR #1128, opened 2020-03-24
DCS-228 Make Eligibility checks available to CA when RO has returned a case to the CA

bell-pa:


server/routes/viewModels/taskListModels.js
```diff
@@ -39,42 +39,45 @@ module.exports = (
...
-    roTasks: getRoTasks,
-    roTasksPostApproval: getRoTasksPostApproval,
-    caTasksEligibility: getCaTasksEligibility,
-    caTasksFinalChecks: getCaTasksFinalChecks,
-    caTasksPostApproval: getCaTasksPostApproval(stage),
-    vary: getVaryTasks({ version, versionDetails, approvedVersion, approvedVersionDetails, licence }),
-  }
+  switch (taskListName) {
+    case 'dmTasks':
+      return getDmTasks({ decisions, tasks, stage, allowedTransition })
```
andrewrlee: Nice! 👍

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -147,7 +145,12 @@ module.exports = {
       return [proposedAddressTask, curfewAddressTask, bassTask, refusalTask].filter(task => task.visible)
     }
 
+    if (!eligible) {
+      return [eligibilityTask, informOffenderTask]
+    }
+
     return [
+      eligibilityTask,
```
andrewrlee: no need for informOffenderTask here?

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -268,13 +271,13 @@ module.exports = {
       visible: validAddress,
     }
 
-    const informOffenderTask = {
+    const invisibleInformOffenderTask = {
       task: 'informOffenderTask',
       visible: true,
```
andrewrlee: If it's invisible why is`visible:true`?

server/routes/viewModels/taskLists/caTasks.js
bell-pa: No, because that task only appears when the offender is Not eligible and that particular line cannot be reached when that is the case.

server/routes/viewModels/taskLists/caTasks.js
bell-pa: I think its a bug in the code.  I created an informOffenderTask at the module level, then discovered this one at the function level so i renamed it.  I don't think it is ever rendered and I suppose I should investigate and remove it.  The Eligibility task post approval and the way it works looks rather broken, but I can’t be be sure of that, no one has complained and  it wasn’t part of the ticket so I ignored it.

PR #1130, opened 2020-03-31
DCS-430 approved form does not render

bell-pa:
This is a fix for a bug introduced by DCS-224


test/services/formService.test.js
```diff
@@ -20,7 +21,7 @@ describe('formService', () => {
 
   beforeEach(() => {
     pdfFormatter = {
-      pickCurfewAddress: jest.fn().mockReturnValue(address),
+      pickCurfewAddress,
     }
```
bell-pa: So that the function call is tested.  Exposes the bug. The test lower down shows that it is fixed.

PR #1136, opened 2020-04-07
Dcs 420 text at the bottom of licence changed to checkboxes

GurnankCheema:


assets/sass/licences-pdf.scss
```diff
@@ -183,7 +183,40 @@ img {
   padding-bottom: 50px;
 }
 
+.extraLargePaddingBottom {
```
brightonsbox, line 186: Is there any reason this needs to be camel case rather than kebab case as the others are?

GurnankCheema: no particular reason .
i'll change it to hyphenated

PR #1138, opened 2020-04-10
DCS-434 moving licences to use dps orb

andrewrlee:


package.json
```diff
@@ -126,6 +126,7 @@
     "pug": "^2.0.4",
     "querystring": "^0.2.0",
     "ramda": "^0.27.0",
+    "request": "^2.88.2",
```
andrewrlee, line 129: this is required as a peer dependency of application insights - was previously being pulled in by a production `node-sass-middleware` dependency, but that has been removed and replaced by a dev dependency of `node-sass`

PR #1144, opened 2020-04-16
DCS-458 moving to deploy typescript application

andrewrlee:


test/data/configClient.test.js
```diff
@@ -1,7 +1,11 @@
+jest.mock('../../server/data/dataAccess/db')
```
andrewrlee, line 1: this would usually be hoisted to the top level . To do that with jest-ts we would have to add babel which bit reluctant to do

server/routes/admin/mailboxes.js
```diff
@@ -49,7 +49,7 @@ module.exports = ({ configClient }) => (router, audited) => {
       return res.redirect(`/admin/mailboxes/edit/${id}`)
     }
 
-    res.redirect('/admin/mailboxes')
+    return res.redirect('/admin/mailboxes')
```
andrewrlee, line 52: typescript warns about inconsistent returns if some places use an explicit return and other places don't

server/services/pdfService.js
```diff
@@ -45,7 +45,7 @@ module.exports = function createPdfService(logger, licenceService, conditionsSer
     const currentTemplateId = getIn(rawLicence, ['licence', 'document', 'template', 'decision'])
 
     if (offenceCommittedBeforeCutoffDecision === offenceCommittedBeforeFeb2015 && templateId === currentTemplateId) {
-      return rawLicence
+      return
```
GurnankCheema, line 48: return null?

server/routes/admin/mailboxes.js
petergphillips: but I didn't think that `res.redirect` returns anything, so `return`ing the value from it looks weird

Dockerfile
```diff
@@ -1,27 +1,41 @@
-FROM node:10-buster-slim
-LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
+# Build stage 1.
+# Create base image with python (required for node-gyp)
 ARG BUILD_NUMBER
 ARG GIT_REF
 
+FROM node:10-buster-slim as base
```
petergphillips, line 6: node 10?  Also don't think it is good to specify `buster` as that is the default anyway so we are pinning for no reason

server/routes/admin/mailboxes.js
andrewrlee: We inherited this style from form app starter in use-of-force.. I think it's a common convention to return `res.redirect` in node apps, e.g:

Screenshot 2020-04-17 at 14 42 34

Dockerfile
andrewrlee: Upgrading from node 10 is on the backlog

server/services/pdfService.js
andrewrlee: if we returned a value from here then we would have to explicitly return a value at the bottom.

GurnankCheema: ok

PR #1149, opened 2020-04-20
DCS-411: CA can view the DMs HDC decision comments

GurnankCheema:
The comments appear only after the DM had made a decision providing the decision is a refusal.
The comments appear in the orange bordered box at the top of the CA tasklist.

### DIFF
--- a/server/utils/licenceStatus.js
+++ b/server/utils/licenceStatus.js
@@ -91,6 +91,7 @@ function getApprovalStageState(licence) {
       dmRefused: approvalRelease.decision === 'No',
       refusalReason,
       bassAccepted,
+      decisionComments: approvalRelease.reasonForDecision,
     },
     tasks: {
       approval: isEmpty(approvalRelease.decision) ? taskStates.UNSTARTED : taskStates.DONE,
--- a/server/utils/licenceStatus.js
+++ b/server/utils/licenceStatus.js
@@ -91,7 +91,7 @@ function getApprovalStageState(licence) {
       dmRefused: approvalRelease.decision === 'No',
       refusalReason,
       bassAccepted,
-      decisionComments: approvalRelease.reasonForDecision,
+      decisionComments: approvalRelease.reasonForDecision ? approvalRelease.reasonForDecision.trim() : null,
     },
     tasks: {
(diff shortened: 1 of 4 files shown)


server/utils/licenceStatus.js
```diff
@@ -176,6 +179,7 @@ function getCaStageState(licence) {
       bassAccepted,
       bassWithdrawn,
       bassWithdrawalReason,
+      caRefused: (approvalRelease.decision === 'Yes' && finalChecksRefusal.decision === 'Yes') || false,
```
bell-pa, line 182: You don't need the `|| false` bit. The value of the  first part of the expression is always either `true` or `false`

server/views/taskList/taskListTemplate.pug
```diff
@@ -23,9 +23,20 @@ block content
   else if licenceStatus.decisions.unsuitableResult
     div.error-banner.marginTopMedium
       strong.bold The offender is presumed unsuitable for HDC release
-  else if user.role != 'DM' && licenceStatus.decisions.refused
+  else if user.role === 'CA' && licenceStatus.decisions.caRefused
     div.error-banner.marginTopMedium
-      strong.bold Home detention curfew refused - #{licenceStatus.decisions.refusalReason}
+      strong.bold Home detention curfew refused by Case Admin - #{licenceStatus.decisions.refusalReason}
+  else if user.role === 'CA' && licenceStatus.decisions.refused && licenceStatus.decisions.decisionComments
```
bell-pa, line 29: Prefer the full name of the role to an abbreviated form: 'Home detention curfew refused by Prison Case Administrator'

server/utils/licenceStatus.js
GurnankCheema: oops, that was me trying code out but then forgot to take it out.
Done

server/views/taskList/taskListTemplate.pug
GurnankCheema: fair point
Done

GurnankCheema: changed to 'prison case admin' and 'decision maker'

PR #1159, opened 2020-04-27
DCS-391 Showing 'inform offender task' when made ineligible in post approval

andrewrlee:
Also fixing wording - the forms are now held in the service so don't need to use nomis


server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -20,7 +20,6 @@ const resubmitToDm = require('./tasks/ca/resubmitToDm')
 const caBlocked = (errorCode) => () => ({ task: 'caBlockedTask', errorCode })
 const eligibilityTask = namedTask('eligibilityTask')
 const eligibilitySummaryTask = namedTask('eligibilitySummaryTask')
-const invisibleInformOffenderTask = namedTask('informOffenderTask')
```
andrewrlee, line 23: it was invisible! It shouldn't have been :-D

PR #1160, opened 2020-04-27
DCS-391 Extracting out bass state calculation

andrewrlee:


server/utils/bassAddressState.js
```diff
@@ -0,0 +1,121 @@
...
+module.exports = { getBassAreaState, getBassState, getBassRequestState }
+
+function getBassRequestState(licence) {
+  const bassRequestAnswer = getIn(licence, ['bassReferral', 'bassRequest', 'bassRequested'])
+  const addressProposedAnswer = getIn(licence, ['proposedAddress', 'addressProposed', 'decision'])
+
+  const bassReferralNeeded = bassRequestAnswer === 'Yes' && addressProposedAnswer === 'No'
+  const bassAreaSpecified = getIn(licence, ['bassReferral', 'bassRequest', 'specificArea']) !== 'No'
+  const bassRequest = getState()
+
```
bell-pa, line 13: getTaskState might be a better name than getState...

andrewrlee: yes - I don't like that it's an inner function either, this is just a whole sale copy over

PR #1162, opened 2020-04-29
DCS-391 simplifying eligibility state calculation

andrewrlee:


server/utils/licenceStatus.js
```diff
@@ -236,10 +224,7 @@ function getEligibilityStageState(licence) {
       addressUnsuitable,
     },
     tasks: {
-      exclusion,
-      crdTime,
-      suitability,
-      eligibility,
+      ...eligibilityState.tasks,
```
andrewrlee, line 227: I could very easily be talked out of this destructuring and instead being more explicit about whats being passed through

server/utils/licenceStatus.js
```diff
@@ -195,12 +195,7 @@ function getCaStageState(licence) {
...
 function getEligibilityStageState(licence) {
-  const { exclusion, excluded } = getExclusionState(licence)
-  const { crdTime, insufficientTime, insufficientTimeContinue, insufficientTimeStop } = getCrdTimeState(licence)
-  const { suitability, unsuitable, unsuitableResult, exceptionalCircumstances } = getSuitabilityState(licence)
-
-  const notEligible = excluded || insufficientTimeStop || unsuitableResult
-  const { eligibility, eligible } = getEligibilityState(notEligible, [exclusion, crdTime, suitability])
+  const eligibilityState = getEligibilityState(licence)
 
   const { curfewAddressApproved, addressReviewFailed } = getCurfewAddressReviewState(licence)
```
andrewrlee, line 200: it winds me up a bit that one of these is a task and one of these is a decision! its not at all obvious from reading it

PR #1163, opened 2020-04-30
DCS-411: changes to HDC refusal message

GurnankCheema:
This is temporary fix for a bug. There may be occasions when a CA only partly completes the HDC refusal task. In these cases the message will display a colon but with nothing to its right hand side. Fix implemented to remove the colon if the CA does not provide a corresponding refusal reason .

### DIFF
--- a/server/views/taskList/taskListTemplate.pug
+++ b/server/views/taskList/taskListTemplate.pug
@@ -23,12 +23,9 @@ block content
   else if licenceStatus.decisions.unsuitableResult
     div.error-banner.marginTopMedium
       strong.bold The offender is presumed unsuitable for HDC release
-  else if user.role === 'CA' && licenceStatus.decisions.caRefused && licenceStatus.decisions.refusalReason
+  else if user.role === 'CA' && licenceStatus.decisions.caRefused
     div.error-banner.marginTopMedium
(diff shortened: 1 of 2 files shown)


server/views/taskList/taskListTemplate.pug
```diff
@@ -23,9 +23,12 @@ block content
   else if licenceStatus.decisions.unsuitableResult
     div.error-banner.marginTopMedium
       strong.bold The offender is presumed unsuitable for HDC release
-  else if user.role === 'CA' && licenceStatus.decisions.caRefused
+  else if user.role === 'CA' && licenceStatus.decisions.caRefused && licenceStatus.decisions.refusalReason
     div.error-banner.marginTopMedium
       strong.bold Home detention curfew refused by prison case admin: #{licenceStatus.decisions.refusalReason}
+  else if user.role === 'CA' && licenceStatus.decisions.caRefused && !licenceStatus.decisions.refusalReason
+    div.error-banner.marginTopMedium
+      strong.bold Home detention curfew refused by prison case admin
```
andrewrlee, line 31: this if/else block has become a bit of a monster. 
Could potentially merge the two cases for ca refused in one through using a ternary operator?

 `#{licenceStatus.decisions.refusalReason ? ': ' +  licenceStatus.decisions.refusalReason : ''}`

PR #1165, opened 2020-05-01
DCS-253 Admin UI for FMB addresses stored in the probation-teams service.

bell-pa:
Merge data from Delius and probation-teams APIs to provide a user interface for adding, updating and removing functional mailbox (FMB) addresses associated with LDUs and Probation Teams. These FMBs are stored in the probation-teams service keyed by Probation Area Code, Local Delivery Unit code and, when attached to a Probation Team, a Probation Team Code.


server/index.js
```diff
@@ -108,6 +109,7 @@ const nomisPushService = createNomisPushService(nomisClientBuilder, signInServic
 const notificationJobs = createNotificationJobs(reminderService, signInService)
 const jobSchedulerService = createJobSchedulerService(dbLockingClient, configClient, notificationJobs)
 const lduService = createLduService(deliusClient, activeLduClient)
+const functionalMailboxService = new FunctionalMailboxService(deliusClient, probationTeamsClient)
 const licenceSearchService = createLicenceSearchService(licenceClient, signInService, nomisClientBuilder)
```
bell-pa, line 113: Um.  Perhaps creating a class was a bad idea - because it is inconsistent, and not really needed. I'm prepared to change this if the reviewer agrees...

server/app.js
```diff
@@ -354,6 +356,10 @@ module.exports = function createApp({
     '/admin/licences/',
     secureRoute(licenceRouter(licenceService, signInService, prisonerService, audit, roNotificationHandler))
   )
+  app.use(
+    '/admin/functionalMailboxes',
+    secureRoute(functionalMailboxRouter(functionalMailboxService), { auditKey: 'FUNCTIONAL_MAILBOX' })
```
andrewrlee, line 361: do you need to add this as an allowed audit key: in audit.js?

server/data/deliusClient.ts
```diff
@@ -79,6 +42,10 @@ module.exports = (signInService) => {
 
       return result.body
     } catch (error) {
+      if (error.status === 404) {
+        logger.debug('Returned 404', path, error.response, error.stack)
+        return undefined
```
andrewrlee, line 47: is there a reason to return undefined rather than null as we do elsewhere in here? 

I don't think debug logging is captured at all in application insights, might be nicer to have a simple info trace, following the pattern as below though possibly with the error stack. 

I think we need to add some context. - "we are calling delius" as the logging isn't name spaced as in java apps
The error.response object is massive which is why we just get the text of the response out below.

server/index.js
andrewrlee: Using classes means that you don't need to define separate interfaces -  you can import the class as a type:

```ts
import  FunctionalMailboxService from '../../..'

export default function someThing(functionalMailboxService: FunctionalMailboxService)
```

server/services/functionalMailboxService.ts
```diff
@@ -0,0 +1,109 @@
...
+  return R.pathOr(
+    '',
+    [0, 'description'],
+    ldus.filter((ldu) => ldu.code === lduCode)
+  )
+}
+
+function mergeProbationTeams(probationTeams: ProbationTeam[], localDeliveryUnitDto: LocalDeliveryUnitDto) {
+  return R.pipe(
+    R.indexBy(R.prop('code')),
```
andrewrlee, line 31: Holy Ramda! 

I've been trying to read this method for a few minutes and still don't really understand what it's doing. 

When playing in a ramda it seem like the ` R.indexBy(R.prop('code')),` is unnecessary as straight away you throw it away be doing the map to extract out the description?

server/app.js
bell-pa: Yes.  You're right. I forgot about auditing. Partly because I couldn't decide where it should be done. (probation-teams or licences).  licences has audit support in place and knows the requesting username, so it would be simple to add it here - which I will do now.

server/data/deliusClient.ts
bell-pa: I think I made this debug level because not found is a normal flow and nothing unusual.  As for undefined vs null - I'll change it.

server/index.js
bell-pa: OK

server/services/functionalMailboxService.ts
```diff
@@ -0,0 +1,109 @@
...
+    R.map(R.pick(['description'])),
+    R.mergeDeepRight(R.propOr({}, 'probationTeams', localDeliveryUnitDto))
+  )(probationTeams)
+}
+
+export function mergeProbationAreaData(probationAreas: ProbationArea[], probationAreaCodes: string[]) {
+  const probationCodeMap = R.pipe(R.indexBy(R.identity), R.map(R.always({})))(probationAreaCodes)
```
andrewrlee, line 38: more idiomatic JS maybe clearer?

```js
const probationCodeMap = probationAreaCodes.reduce((result, code) => ({...result, [code]: {}}) , {})

const probationAreaMap = probationAreas.reduce((result, {code, description}) => ({...result, [code]: { description }}), {})
```

PR #1170, opened 2020-05-06
DCS-476 Showing reason when CA refuses case before DM handover

andrewrlee:
Also fixing refuse button - it previously never changed value even if refused
This is because it was looking at the wrong decision

Tidied up the logic for displaying warning message too.


server/utils/licenceStatus.js
```diff
@@ -182,7 +182,8 @@ function getCaStageState(licence) {
       bassAccepted,
       bassWithdrawn,
       bassWithdrawalReason,
-      caRefused: approvalRelease.decision === 'Yes' && finalChecksRefusal.decision === 'Yes',
+      caRefused: finalChecksRefusal.decision === 'Yes',
+      refusalReason,
```
andrewrlee, line 186: this is being declared in `caStageState`  - refusalReason will be overridden in post approval stages to be DM reason or if not exisitng CA reason. There is a test that covers this.

PR #1173, opened 2020-05-06
DCS-253 Bug fix

bell-pa:
A community API request for probation teams can return one of:
- A 200 response and a JSON body containing an object having a "content" field that is an array of result objects
- A 200 response and a JSON body containing an object that does not have a "content" field
- A 404 response

Requests for LDUs return either
- A 200 response and a JSON body containing an object having a "content" field that is an array of result objects
- A 404 response

These cases need handling correctly.

### DIFF
--- a/server/services/functionalMailboxService.ts
+++ b/server/services/functionalMailboxService.ts
@@ -63,7 +63,7 @@ export class FunctionalMailboxService {
   }
 
   getLdusForProbationArea = async (probationAreaCode): Promise<LduMap> => {
-    const [{ content: ldus = [] } = {}, { localDeliveryUnits = {} } = {}] = await Promise.all([
+    const [{ content: ldus }, { localDeliveryUnits = {} } = {}] = await Promise.all([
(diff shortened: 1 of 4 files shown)


server/services/functionalMailboxService.ts
```diff
@@ -55,7 +55,7 @@ export class FunctionalMailboxService {
   ) {}
 
   getAllProbationAreas = async () => {
-    const [{ content: probationAreas = [] } = {}, probationAreaCodes = []] = await Promise.all([
+    const [{ content: probationAreas } = { content: [] }, probationAreaCodes = []] = await Promise.all([
```
andrewrlee, line 58: meh.. would have been tempted to push quirks of the dependent service down in to client - so the client just always returns a list.

I'm going to have a quick look at fixing in community-api but doesn't need to hold up this PR.

andrewrlee: actually this looks a little weird - the client for `getAllTeamsForLdu` does handle the quirks in the service layer but `getAllProbationAreas` doesn't and has the defaulting at this level?

PR #1181, opened 2020-05-26
DCS-493 HDC - Don't log confidential headers

bell-pa:


server/data/restClientBuilder.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export const constantTokenSource = (token): TokenSource => async () => {
+  if (!token) {
+    throw Error('Unauthorised access')
+  }
+  return token
+}
+
+export const dynamicTokenSource = (signInService, oauthServiceName: string): TokenSource => async () => {
```
andrewrlee, line 36: could possibly call this something about client credentials / anonymous / system to system token?

server/data/restClientBuilder.ts
```diff
@@ -0,0 +1,137 @@
...
+          .get(`${apiUrl}${path}`)
+          .query(query)
+          .agent(keepaliveAgent)
+          .set('Authorization', `Bearer ${token}`)
+          .set(headers)
+          .retry(2, (err, res) => {
+            if (res) {
+              if (res.status >= 300 && res.status !== NOT_FOUND && res.status !== UNAUTHORIZED) {
+                logger.warn(
+                  `Retry handler found API error with status: '${res.status}', message: '${res.error?.message}'`
```
andrewrlee, line 70: probably don't want to retry on any 4XX client error?

server/data/nomisClientBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+    getUserInfo(userName) {
+      return oauthRestClient.getResource(`/api/user/${userName}`)
+    },
+
+    getLoggedInUserInfo(): Promise<Profile> {
+      return oauthRestClient.getResource(`/api/user/me`)
+    },
+
+    getUserRoles(): Promise<Role> {
+      return oauthRestClient.getResource(`/api/user/me/roles`)
```
andrewrlee, line 179: Out of scope for this story but is there anything to stop us from splitting out the auth calls into a separate client?

server/data/restClientBuilder.ts
bell-pa: Done

server/data/restClientBuilder.ts
bell-pa: Yes, of course! No idea why I did this.  Now retrying on res.status >= 500.

PR #1182, opened 2020-06-01
DCS-471 CAs HDC refusal is not recorded if reason not provided

GurnankCheema:
If a CA refuses HDC by selecting Yes in the page /hdc/finalChecks/refuse/{bookingId} but doesn't follow it up with a reason in page /hdc/finalChecks/refusal/{bookingId}, then when CA returns back to the CA tasklist, the previous Yes decision is deleted because the refusal has not been fully completed. The refusal can not be treated as complete without a reason being registered. 
The tasklist will not therefore display any refusal message and will have returned to the original state.

### DIFF
--- a/server/routes/eligibility.js
+++ b/server/routes/eligibility.js
@@ -1,6 +1,6 @@
 const formConfig = require('./config/eligibility')
 const { asyncMiddleware } = require('../utils/middleware')
-const { getIn, firstItem } = require('../utils/functionalHelpers')
+const { getIn } = require('../utils/functionalHelpers')
 const createStandardRoutes = require('./routeWorkers/standard')
 
 module.exports = ({ licenceService, nomisPushService }) => (router, audited, config) => {
--- a/server/routes/taskList.js
+++ b/server/routes/taskList.js
@@ -7,7 +7,7 @@ const { asyncMiddleware } = require('../utils/middleware')
 const { getLicenceStatus } = require('../utils/licenceStatus')
 const { getStatusLabel } = require('../utils/licenceStatusLabels')
 
-const { isEmpty } = require('../utils/functionalHelpers')
+const { isEmpty, getIn } = require('../utils/functionalHelpers')
 const getTaskListModel = require('./viewModels/taskListModels')
 const logger = require('../../log')
 const { getTasksForBlocked } = require('./viewModels/taskLists/caTasks')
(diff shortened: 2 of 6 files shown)


server/routes/eligibility.js
```diff
@@ -1,6 +1,6 @@
 const formConfig = require('./config/eligibility')
 const { asyncMiddleware } = require('../utils/middleware')
-const { getIn, firstItem } = require('../utils/functionalHelpers')
```
GurnankCheema, line 3: unused variable

server/views/taskList/taskListTemplate.pug
```diff
@@ -26,7 +26,7 @@ block content
       strong.bold The offender is presumed unsuitable for HDC release
   else if user.role === 'CA' && decisions.caRefused
     div.error-banner.marginTopMedium
-      strong.bold Home detention curfew refused by prison case admin#{decisions.refusalReason ?': ' + decisions.refusalReason : ''}
```
GurnankCheema, line 29: line 29 with the ternary was a temporary fix in a previous ticket. Ternary no longer needed.

server/routes/taskList.js
```diff
@@ -111,9 +111,20 @@ module.exports = ({ prisonerService, licenceService, audit, caService }) => (rou
         return res.redirect(`/hdc/review/licence/${bookingId}`)
       }
 
-      const licenceStatus = getLicenceStatus(licence)
+      let licenceStatus = getLicenceStatus(licence)
       const { statusLabel } = getStatusLabel(licenceStatus, req.user.role)
 
+      if (req.user.role === 'CA' && licence && licenceStatus.stage !== 'ELIGIBILITY') {
```
andrewrlee, line 117: Does this code not seem a bit out of place? 
it seems very specific for such a generic function - handling edge cases like this in generic code is a code smell.
Lots of these kinds of changes mount up and make understanding what is going on much more difficult.

andrewrlee: I think there might be a few possible alternatives to handling this a bit better, let's talk about it after lunch

server/views/taskList/taskListTemplate.pug
andrewrlee: you won't be able to remove this ternary - as there will still be lots of data in the database in the bad state. We can only fix it going forward

GurnankCheema:
different approach taken,  on a new branch

PR #1187, opened 2020-06-03
DCS-502 Create Typescript interface(s) that describe the structure of licences.licence database objects.

bell-pa:


server/services/licenceService.ts
```diff
@@ -22,39 +24,54 @@ const {
...
+        return null
+      }
 
-      const licence = getIn(rawLicence, ['licence'])
+      const { licence, stage } = rawLicence
       if (!licence) {
         return null
       }
-      const formattedLicence = formatObjectForView(licence)
-      const approvedVersionDetails = rawVersionDetails ? formatObjectForView(rawVersionDetails) : {}
```
bell-pa, line 50: I can't see how {formatObjectForView} could change {licence} or {rawVersionDetails} so I've removed both invocations.  All the tests are passing, I've tried checking the code for the field names used within that function and haven't found any matches.  

Is this risky???

server/services/licenceService.ts
```diff
@@ -1,9 +1,11 @@
-const moment = require('moment')
```
bell-pa, line 1: This class / file is too big, but I'll leave dismembering it for another time.

server/data/licenceClient.ts
```diff
@@ -119,8 +151,9 @@ module.exports = {
     }
 
     return db.query(query)
-  },
+  }
 
+  // eslint-disable-next-line class-methods-use-this
```
andrewrlee, line 156: would be tempted to move to inject `db` in the future, then we can use normal mocking in the db tests and not have these warnings to suppress. It's a bit random we actually require it directly

server/data/licenceClientTypes.ts
```diff
@@ -0,0 +1,24 @@
+import { Licence } from './licenceTypes'
+
+export interface Case {
+  licence: Licence
+  booking_id: number
```
andrewrlee, line 5: If we get more TS coverage we could map these names using a column alias to camelCase

server/data/licenceClient.ts
```diff
@@ -41,12 +60,13 @@ module.exports = {
       return rows[0]
     }
 
-    return {}
-  },
+    return undefined
```
andrewrlee, line 63: I can only see this being used in one place, is that right?

PR #1188, opened 2020-06-04
DCS-506 Move a few classes over to typescript

andrewrlee:


.eslintrc.json
```diff
@@ -19,6 +19,7 @@
         "project": "./tsconfig.json"
       },
       "rules": {
+        "@typescript-eslint/no-use-before-define": 0,
```
andrewrlee, line 22: we disable this rule for JS - it's generally recommend to disable for typescript and easier to convert modules if follow similar rules to JS linting rules

PR #1193, opened 2020-06-09
Dcs 471 prevent ca refusing hdc without providing reason

GurnankCheema:
http://localhost:3000/hdc/finalChecks/refuse/{bookingId}

The two screens (refuse and refusal) have been merged into one (finalChecks/refuse).
The finalChecks/refuse/ screen uses a customised route handler for POSTS.
It  ensures the user inputs are validated before being persisted to DB preventing partial data being persisted.
The data is sanitised to prevent any 'out of time reasons checks' being persisted unless the parent 'out of time' radio is also selected


server/routes/config/finalChecks.js
```diff
@@ -91,22 +91,6 @@ module.exports = {
     pageDataMap: ['licence', 'finalChecks', 'refusal'],
     saveSection: ['finalChecks', 'refusal'],
     fields: [{ decision: {} }, { reason: {} }, { outOfTimeReasons: {} }],
```
andrewrlee, line 93: I wonder if you delete the entire refuse section?

PR #1194, opened 2020-06-10
DCS-498 Add loadLicences.js and saveLicences.js scripts to package.json

bell-pa:
Now they can be added to the PATH using 'npm link'

### DIFF
--- a/server/saveLicences.js
+++ b/server/saveLicences.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 const fs = require('fs').promises
 const { Client } = require('pg')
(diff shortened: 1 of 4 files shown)


server/saveLicences.js
```diff
@@ -34,6 +34,13 @@ const writeToFile = async (json) => {
   return fh.close()
 }
 
+/*
+const writeToStdOut = async (json) => {
+  process.stdout.write(json)
+  process.stdout.end()
+}
+*/
```
andrewrlee, line 42: bit of a comment here

PR #1201, opened 2020-06-26
DCS-317 Read-only access for prison staff who do not have the CA, DM or RO role.

bell-pa:
Show the case list and a  read-only view of a case for any authenticated user who has the `PRISON` role but none of the `LICENCE_DM`, `LICENCE_CA` or `LICENCE_RO` roles.

Also extended the `collectAuditStatistics.js `script to include a summary of outcomes for each node in the tree of user journeys. This does not affect affect any production code or tests so feel free to ignore this and changes to files in `server/statistics` where most of scripts functionality resides.


server/utils/Result.ts
```diff
@@ -1,8 +1,15 @@
...
+  fail(): E
+  map<T>(f: (v: S) => T): Result<T, E>
+  mapAsync<T>(f: (v: S) => Promise<T>): Promise<Result<T, E>>
+  orRecoverAsync(f: () => Promise<Result<S, E>>): Promise<Result<S, E>>
+  flatMap<T>(f: (v: S) => Result<T, E>): Result<T, E>
+  match<T>(successFunction: (v: S) => T, failFunction: (v: E) => T): T
+  isSuccess(): boolean
+}
+
+export const Success = <S, E>(value: S): Result<S, E> => ({
```
andrewrlee, line 12: nice

server/services/caseListService.ts
```diff
@@ -1,110 +1,84 @@
...
+      case 'PRISON':
+      case 'CA':
+      case 'DM':
+        return getPrisonCaseList(token)
+
+      case 'RO':
+        return getROCaseList(username, token)
+
+      default:
+        throw Error(`Illegal state: unrecognised role ${role}`)
```
andrewrlee, line 29: nice!

PR #1202, opened 2020-06-28
DCS-369: CA can view additional conditions in conditions summary page

GurnankCheema:
3 new sections added to the /hdc/licenceConditions/standard/{bookingId} that are view only data, so that CA can see the additional, PSS and bespoke conditions without having to navigate to the conditions selection page or the printed licence.

looks like : 

img1

img2


server/views/licenceConditions/standard.pug
```diff
@@ -3,6 +3,7 @@ extends ../layout
 block content
 
   - var additionalConditionsValue = data.additionalConditionsRequired || null;
+  - var licenceConditions = licenceWithReconstructedConditions.licenceConditions || {}
```
andrewrlee, line 6: this seems a bit confusing, why do we have reference to 2 things here - `conditions` and `standardConditons`?
Can we simplify this?

server/views/licenceConditions/standard.pug
```diff
@@ -19,6 +20,52 @@ block content
...
+                    span #{item.variable}
+    if results === false
+      | No additional conditions selected
+
+  h2.heading-large Post sentence supervision conditions
+  div.pure-u-1.pure-u-md-2-3
+    - var results = false
+    ol.numberedList
+      for condition in licenceConditions
+          if condition.group === 'Post-sentence supervision only' && condition.group === 'Post-sentence supervision only'
```
andrewrlee, line 44: doubles up condition?

server/views/licenceConditions/standard.pug
```diff
@@ -19,6 +20,52 @@ block content
         for condition in conditions
           li.midPaddingBottom #{condition.text}
 
+
+  h2.heading-large Additional conditions
+  div.pure-u-1.pure-u-md-2-3
+    - var results = false
+    ol.numberedList
+      for condition in licenceConditions
+          if condition.group !== undefined && condition.group !== 'Post-sentence supervision only' && condition.group !== 'Bespoke'
```
andrewrlee, line 29: unnecessary undefined check?
Might be nicer to move this logic out of the template? - passing the types of conditions down as separate collections..

server/views/licenceConditions/standard.pug
andrewrlee: it seems like the variable called `conditions` now needs a better name?

andrewrlee: but also do you even need to pass it in? does licenceConditions also include it?

server/views/licenceConditions/standard.pug
GurnankCheema: oops, I was tinkering with mixins that took in 2 args because the additional and  PSS conditions needed them. Then I ditched the idea of a mixin but forgot to take out the above.
Done.

server/views/licenceConditions/standard.pug
GurnankCheema: The undefined check was necessary because the user could continue to subsequent pages even if they didn't select yes or no 
Without checking for undefined would have meant the user seeing a bespoke condition that wasn't explicitly approved. 
Re 'passing the types of conditions down as separate collections..' I started down that path but decided against it because other pages have used the approach of sending everything to the front

server/views/licenceConditions/standard.pug
GurnankCheema: the standard conditions were already there. 
This ticket was to add the additional, pss and bespoke conditions

GurnankCheema: amended

server/views/licenceConditions/standard.pug
GurnankCheema: amended

server/services/conditionsService.js
```diff
@@ -336,3 +362,16 @@ function getSubmissionForCondition(conditionId, inputtedConditions) {
...
+
+function constructConditionTexts(conditions) {
+  const formattedConditions = []
+  conditions.forEach((condition) => {
+    let contentText = ''
+    condition.forEach((conditionObject) => {
+      contentText += conditionObject.text || conditionObject.variable
+    })
+    formattedConditions.push({ text: contentText })
+  })
```
andrewrlee, line 374: code without mutation can be easier to read and reason about.
can you think how to refactor this to use map and join?

server/services/conditionsService.js
```diff
@@ -199,6 +199,31 @@ module.exports = function createConditionsService({ use2019Conditions }) {
...
+    if (licenceConditions.length > 0) {
+      licenceConditions.forEach((condition) => {
+        if (condition.group === 'Bespoke' && condition.approved === 'Yes') {
+          bespokeConditionsContent.push(condition.content)
+        } else if (condition.group === 'Post-sentence supervision only') {
+          pssConditionsContent.push(condition.content)
+        } else if (condition.group !== 'Bespoke' || condition.group === undefined) {
+          additionalConditionsContent.push(condition.content)
+        }
+      })
```
andrewrlee, line 217: is the `licenceConditions.length` check required?

can you think how to simplify this and refactor it to use filter?

GurnankCheema: the length check isn't needed
can use filter but the licenceConditions isn't always an array 
It is an object if no conditions are selected. This would then need to be converted to an array so .filter can be used. Something like this ...
   ``` if (typeof licenceConditions === 'object') {
      licenceConditions = Array.from(licenceConditions)
    }```

server/services/conditionsService.js
GurnankCheema: amended

.gitignore
```diff
@@ -29,3 +29,4 @@ out
 .gradle
 test-results/
 licences\.iml
+licence-specs/src/*.iml
```
andrewrlee, line 32: this is fine - an alternative would be to merge this with line above with `**/*.iml`

PR #1205, opened 2020-07-08
DCS-564 whitelist -> allow_list

andrewrlee:


helm_deploy/values-dev.yaml
```diff
@@ -10,7 +10,7 @@ image:
 
 ingress:
   enabled: true
-  enable_whitelist: false
+  enable_allowlist: false
```
matt-crowson-moj, line 13: should this be `enable_allow_list`?  Same applies elsewhere...

andrewrlee: fixed this

PR #1221, opened 2020-08-19
DCS-609 Adding fallback to load RO caseload

andrewrlee:
Currently to build up the RO caselist, we look up offender data using a staff code

1.) We initially search the local staff table for a mapping from auth user to delius staff code
2.) If that fails, we make a request to delius to find the staff code for the currently logged in user.

After the probation area split on the weekend, new staff codes will be generated so our mappings in 1.) will
become out of date. We hoped to migrate these users to use delius credentials to access these services but
unfortunately have not had time to organise comms.

To get round this, this change will store the users currently mapped delius username in the local db.
It will then use this as a fallback when 1.) and 2.) fail to allow retrieval of an appropriate staff code
which can then be used to retrieve additional data.

### DIFF
--- a/migrations/20200818182127_add-delius-username-to-staff-id.js
+++ b/migrations/20200818182127_add-delius-username-to-staff-id.js
@@ -1,0 +1,13 @@
+exports.up = async (knex) => {
+  await knex.schema.alterTable('staff_ids', (table) => {
+    table.string('delius_username').nullable()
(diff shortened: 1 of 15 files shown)


JamieDraperUK (approved):
lgtm 👍

bell-pa (approved):
LGTM

PR #1222, opened 2020-08-19
DCS-609 Use delius username to link to staff data from local store 

andrewrlee:
when retrieving contact details


GurnankCheema (approved):
LGTM, I think

PR #1223, opened 2020-08-20
DCS-590 :bug: :white_check_mark:  BASS request + Approved Premises problem

bell-pa:
:bug:
Prison submitted BASS area to RO.
RO updated to AP required and had added the AP address.
Only the original BASS area address is visible to the CA at the prison when expecting to see Approved Premises address.

:white_check_mark: 
Fixed this bug and added some tests to guard against regressions. 
All tests and integration tests pass but test coverage is poor to non-existent for BASS + Approved Premises.  
This change might have broken some other part of the application but I have tested the main flows in this area by hand and they all seem correct.


andrewrlee (approved):
LGTM

PR #1224, opened 2020-08-20
DCS-613 Adding screens to allow app support to manage users HDC specific delius roles

andrewrlee:
App support need to be able to manage HDC specific delius roles. 

These screens allow an admin staff to search for delius users by username and see what HDC specific roles they have.
It also allows them to add HDC roles.

Haven't mapped the readonly role as we aren't using it yet.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -29,6 +29,7 @@ const defaultRouter = require('./routes/default')
 
 const adminRouter = require('./routes/admin/admin')
 const userAdminRouter = require('./routes/admin/users')
+const manageRolesRouter = require('./routes/admin/manageRoles')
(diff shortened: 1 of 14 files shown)


server/app.js
```diff
@@ -347,8 +348,9 @@ module.exports = function createApp({
   app.use('/admin/', secureRoute(adminRouter()))
   app.use(
     '/admin/roUsers/',
-    secureRoute(userAdminRouter({ userAdminService, signInService, migrationService }), { auditKey: 'MIGRATION' })
```
andrewrlee, line 350: I've reverted this back to what it was before! didn't really make sense adding a MIGRATION specific key

PR #1227, opened 2020-09-01
DCS-480 

bell-pa:
Bug: Postpone an approved licence. CA Caselist status is now 'Licence created' and Create Licence task is available.
Case list status is now correct for CA and DM.  DM 'action' also changed to 'View' from 'Change'.


GurnankCheema (approved):
LGTM

PR #1228, opened 2020-09-03
DCS-604 Prevent logging of post requests in licences

bell-pa:
Stop using the Winston Console transport when application insights is used.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -23,6 +23,7 @@ const config = require('./config')
 const healthFactory = require('./services/healthcheck')
 
 const logger = require('../log.js')
+const appInsights = require('../azure-appinsights')
 const auth = require('./authentication/auth')
 
 const defaultRouter = require('./routes/default')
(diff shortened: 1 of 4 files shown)


server/app.js
```diff
@@ -200,29 +201,31 @@ module.exports = function createApp({
...
-  // Request logging
-  app.use(
-    expressWinston.logger({
-      winstonInstance: logger,
-      meta: true,
-      dynamicMeta(req, res) {
-        const meta = {
-          userEmail: req.user ? req.user.username : null,
-          requestId: req.id,
-          sessionTag: req.user ? req.user.sessionTag : null,
```
andrewrlee, line 212: don't think this could have been working?!

andrewrlee (approved):
LGTM

server/app.js
bell-pa: I've removed it completely and replaced with middleware that logs path, username and role as custom dimensions.

andrewrlee (approved):
LGTM

PR #1231, opened 2020-09-15
DCS-599 Script to disable / enable auth service users.

bell-pa:
Read a list of auth service usernames from stdin, one username per line.
 Disable or enable the account for every username in the list, printing the resulting user state to stdout as JSON.
This script occurs twice, once as disableUsers.ts and again (via a soft link) as enableUsers.ts
The script chooses whether to enable or disable users from the value of process.argv[1]

 If this script is included in a Docker image (as it will be if deployed to the cluster environments)
 it can be run in one of the deployed pods, with input redirected from a local file. That is, a file
 hosted on the machine where the kubectl command is run. Running the script in a pod means that the
 script will see correctly configured environment variables for that namespace.

 Such an invocation might look like:
```bash
 kubectl -n licences-prod exec -i licences-<pod_id> -- node server/disableUsers.js < usernames.txt
```
 Where `pod_id` is obtained from a previous `kubectl get pods` command and `usernames.txt` is a file in the current
 directory which contains auth service usernames, one per line.

### DIFF
--- a/server/disableUsers.ts
+++ b/server/disableUsers.ts
@@ -24,6 +24,8 @@ import { buildRestClient, clientCredentialsTokenSource, TokenSource } from './da
 import createSignInService from './authentication/signInService'
 import config from './config'
 
+const writeStdOut = (data) => new Promise((resolve) => process.stdout.write(data, () => resolve()))
+
 /**
  * Always return the first token retrieved from the wrapped TokenSource.
  * Sufficient for a script that will complete within the cached token's expiry time.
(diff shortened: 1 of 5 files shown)


andrewrlee (dismissed):
LGTM

server/disableUsers.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const updateUsers = (usernames: string[], disableUsers = true) =>
+  usernames.reduce(
+    (promise, username) =>
+      promise
+        .then(() => updateUser(username, disableUsers))
+        .then((data) => new Promise((resolve) => process.stdout.write(`${JSON.stringify(data)}\n`, () => resolve()))),
+    Promise.resolve()
+  )
```
andrewrlee, line 73: meh.. kind of think the imperative approach would be easier to read and reason about..

```js
// eslint-disable-next-line no-restricted-syntax
// eslint-disable-next-line no-await-in-loop
...
const log = text => new Promise((resolve) => process.stdout.write(text, () => resolve()))
          
for (const user of users) {
       const data = await updateUser(username, disableUsers)
       await log(`${JSON.stringify(data)}\n`)
}

await log(`Done\n`)
process.exit(0)
```

server/disableUsers.ts
```diff
@@ -0,0 +1,83 @@
...
+const enableAuthUser = (username) => oauthRestClient.putResource(`/api/authuser/${username}/enable`, {})
+const getUser = (username) => oauthRestClient.getResource(`/api/authuser/${username}`)
+
+const updateUser = async (username, disableUser: boolean) => {
+  if (username) {
+    try {
+      await (disableUser ? disableAuthUser(username) : enableAuthUser(username))
+      return getUser(username)
+    } catch (error) {
+      // Carry on...
```
andrewrlee, line 60: possibly worth logging an error?

server/disableUsers.ts
```diff
@@ -0,0 +1,83 @@
...
+ * The script chooses whether to enable or disable users from the value of process.argv[1]
+ *
+ * If this script is included in a Docker image (as it will be if deployed to the cluster environments)
+ * it can be run in one of the deployed pods, with input redirected from a local file. That is, a file
+ * hosted on the machine where the kubectl command is run. Running the script in a pod means that the
+ * script will see correctly configured environment variables for that namespace.
+ *
+ * Such an invocation might look like:
+ *
+ * kubectl -n licences-prod exec -i licences-<pod_id> -- node server/disableUsers.js < usernames.txt
```
andrewrlee, line 16: this is nice

server/disableUsers.ts
bell-pa: Done

andrewrlee (approved):
LGTM

PR #1232, opened 2020-09-25
DCS-632 Switch to using staffId based instead of staffCode based community-api endpoints

bell-pa:
🚧   Add staffIdentifier column to staff_ids table. Preparatory work to be deployed to prod.


petergphillips (approved):
LGTM

PR #1234, opened 2020-09-29

bell-pa:


server/data/licenceClient.ts
```diff
@@ -134,7 +134,7 @@ export class LicenceClient {
   async getDeliusIds(nomisUserName): Promise<DeliusIds[]> {
     const query = {
       text:
-        'select staff_id "staffCode", delius_username "deliusUsername" from v_staff_ids where upper(nomis_id) = upper($1)',
+        'select staff_id "staffIdentifier", delius_username "deliusUsername" from v_staff_ids where upper(nomis_id) = upper($1)',
```
andrewrlee, line 137: This looks like its reading the column that stores staff code as staffIdentifier? 
Isn't there a new column to read from?

andrewrlee: Should this be `staff_identifier "staffIdentifier"` ?

bell-pa: Doh.  Fixed now.

test/mockClients.ts
```diff
@@ -1,69 +1,7 @@
...
-export interface DeliusClientMock extends DeliusClient {
-  getStaffDetailsByStaffCode: jest.Mock<Promise<StaffDetails>, [string]>
-  getStaffDetailsByUsername: jest.Mock<Promise<StaffDetails>, [string]>
-  getROPrisoners: jest.Mock<Promise<any>, [string]>
-  getAllOffenderManagers: jest.Mock<Promise<Array<CommunityOrPrisonOffenderManager>>, [string]>
-  getAllProbationAreas: jest.Mock<Promise<Page<ProbationArea>>, []>
-  getAllLdusForProbationArea: jest.Mock<Promise<Page<Ldu>>, [string]>
-  getAllTeamsForLdu: jest.Mock<Promise<Page<ProbationTeam>>, [string, string]>
-  addResponsibleOfficerRole: jest.Mock<Promise<void>, [string]>
-}
```
andrewrlee, line 30: 🥳

test/services/migrationService.test.ts
```diff
@@ -1,8 +1,12 @@
+import { mocked } from 'ts-jest/utils'
 import MigrationService, { Flag } from '../../server/services/migrationService'
 import { delius } from '../../server/config'
+import { DeliusClient, StaffDetails } from '../../server/data/deliusClient'
+
+jest.mock('../../server/data/deliusClient')
 
 describe('MigrationService', () => {
-  let deliusClient
+  let deliusClient: DeliusClient
```
andrewrlee, line 9: Not sure but think this could be simplified by declaring as: `deliusClient: jest.Mocked<DeliusClient>` ? 
then you would only need to use `mocked` once?

`deliusClient = mocked(new DeliusClient(undefined))`

rather than wrapping everytime you want to do an assertion..

types/probationTeams.d.ts
```diff
@@ -1,73 +1,23 @@
...
-  probationTeams: ProbationTeamMap
-}
-
-export interface ProbationTeamMap {
-  [code: string]: {
-    description?: string
-    functionalMailbox?: string
-  }
-}
+// export interface LduMap {
```
andrewrlee, line 1: commented code..

andrewrlee (approved):

PR #1235, opened 2020-09-30

bell-pa:


bell-pa:
When retrieving staff details from Delius fall back to using staff code if no staff identifier found in `staff_ids` table.
Display `staffIdentifier` in migrate view for user.

andrewrlee (approved):
LGTM

PR #1236, opened 2020-10-01

bell-pa:
* Fixes mistake in handling response to community-api end-point `GET /offenders/nomsNumber/${offenderNo}/allOffenderManagers` which returns `staffId` not `staffIdentifier`
* Show staff identifier in RO details page linked from offender details section of task list
* Admin changes to allow search by staff identifier instead of staff code on Delius api page.
* Show staff identifier in messages and RO related information


andrewrlee (approved):
LGTM

PR #1237, opened 2020-10-06
DCS-359 Agency Notification form displays Actual Release Date

GurnankCheema:
The form now displays the Actual Release Date as well as the Conditional Release Date.
As a side issue, a change was also made to mock-server/routes/delius.js as the logic to match the RO now uses the unique staffId rather than the non-unique staffCode value.

Agency notification form before changes
=============================

Screenshot 2020-10-06 at 09 35 13

Agency notification form after changes
=============================
The relevant change is to the line "The following prisoner is due to be released on Home Detention Curfew on 13thSeptember 2019"
-------------------------

Screenshot 2020-10-06 at 11 06 08


bell-pa (approved):
LGTM

PR #1239, opened 2020-10-06
DCS-631 removed inappropriate warning message

GurnankCheema:
The ticket was for the removal of the warning message **“If any of the above details have been changed please make sure the prison case admin is notified to ensure a new licence is created if applicable”**. 

The warning message was appearing in the screen seen by the RO when attempting to vary a licence’s conditions. A vary can only be done once a prisoner has been released. The message was not appearing at any stage of licence creation prior to the prisoner being released. 
Once a prisoner has left prison, the only official thats deal with a prisoner is an RO. CAs and DMs are no longer relevant because they only deal with in-prison cases

The effect of the above is, the message shouldn’t be displayed at all, either before or after prison release. So I have deleted server/views/includes/roModificationWarning.pug (as it was the small file that defined the warning message) and deleted any references to it application-wide.


bell-pa (approved):
LGTM

PR #1243, opened 2020-10-27
DCS-427 HDC - Address change following Licence Variation update

bell-pa:
The updated address was not saving into the new version of the Licence when generated.
The vary address task only ever worked when a pre-existing licence object contained just a proposed address.  It failed to take account of objects where a bass address or approved premises address (for the bass or curfew address) had been specified.

### DIFF
--- a/server/routes/vary.ts
+++ b/server/routes/vary.ts
@@ -1,10 +1,11 @@
-/**
- * @typedef {import("../services/prisonerService").PrisonerService} PrisonerService
- */
-const { asyncMiddleware } = require('../utils/middleware')
-const createStandardRoutes = require('./routeWorkers/standard')
-const { pickBy, getFieldName, isEmpty, firstItem, getIn, lastItem } = require('../utils/functionalHelpers')
-const formConfig = require('./config/vary')
+import { asyncMiddleware } from '../utils/middleware'
+import createStandardRoutes from './routeWorkers/standard'
+import { firstItem, getFieldName, getIn, isEmpty, lastItem, pickBy, selectPathsFrom } from '../utils/functionalHelpers'
+import formConfig from './config/vary'
+import { LicenceService } from '../services/licenceService'
+import { PrisonerService } from '../../types/licences'
+import { Licence } from '../data/licenceTypes'
+import { pickCurfewAddress } from '../services/utils/pdfFormatter'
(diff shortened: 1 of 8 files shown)


server/services/licenceService.ts
```diff
@@ -485,8 +505,9 @@ export class LicenceService {
```
bell-pa, line 505: This is where the vary address form data is used to update the licence.  Before doing that generate a version of fieldConfig that selects the address use by pdfFormatter

server/routes/vary.ts
```diff
@@ -19,19 +20,37 @@ const expectedFieldsForForm = {
...
- * @param {object} args
- * @param {any} args.licenceService
- * @param {PrisonerService} args.prisonerService
+ * The data for the varyAddress form can come from one of four locations within the licence object.
+ * The pickCurfewAddress function from the pdfFormatter module returns the address that will be printed on a PDF licence.
  */
-module.exports = ({ licenceService, prisonerService }) => (router, audited) => {
+function extractFormData(formName: string, licence: Licence, licencePosition: string[]) {
+  if (formName === 'varyAddress') {
+    return pickCurfewAddress(selectPathsFrom(licence))
```
bell-pa, line 28: Display the address shown on the licence pdf .

server/services/licenceService.ts
```diff
@@ -38,6 +42,22 @@ export interface LicenceRecord {
   approvedVersionDetails: ApprovedVersionDetails
 }
 
+export function adaptFieldConfigToSelectWorkingAddress(existingLicence, fieldConfigs) {
+  const curfewAddressPath = pickCurfewAddressPath(selectPathsFrom(existingLicence))
+
+  const modifiedFieldConfigs = R.clone(fieldConfigs).map((fieldConfig) => {
+    const fieldName = Object.keys(fieldConfig)[0]
+    const { licencePosition } = fieldConfig[fieldName]
+    if (licencePosition[0] === 'proposedAddress' && licencePosition[1] === 'curfewAddress') {
```
andrewrlee, line 51: guessing the elements in these don't refer to any thing other than positions in a path? (we cant extract them out and give them a better name?)

andrewrlee (approved):
LGTM

PR #1244, opened 2020-10-28
DCS-650 Bug fix

bell-pa:
CA specifies Bass address after rejected curfew address, RO specifies approved premises, CA can't see this.
Fixed for this particular case by testing the relevant form values for equality with 'Yes' instead of for truthy.


andrewrlee (approved):
LGTM

PR #1247, opened 2020-11-17
DCS-713 DM should not be shown Postpone task.

bell-pa:
Removed postpone task from DM tasklist.


package.json
```diff
@@ -166,7 +166,7 @@
     "jest-html-reporter": "3.3.0",
     "jest-junit": "^11.1.0",
     "lint-staged": "10.5.0",
-    "nock": "^13.0.2",
```
andrewrlee, line 169: Looks like this is limited to not update past the current major version (typescript is stuck at 3.x)

tend to use `npx npm-check-updates -u` to update to the latest.

andrewrlee (approved):
LGTM

PR #1254, opened 2021-01-08
DCS-775 Fix error handling on pdf generation. To prevent from hanging…

andrewrlee:
… forever

### DIFF
--- a/server/routes/forms.js
+++ b/server/routes/forms.js
@@ -85,11 +85,12 @@ module.exports = ({ formService }) => (router) => {
           { ...pageData, ...curfewData, port },
           { filename, pdfOptions }
         )
+
         logger.info(`Returning rendered PDF for form '${templateName}'`)
         return pdf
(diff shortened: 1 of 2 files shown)


server/routes/forms.js
```diff
@@ -85,11 +85,12 @@ module.exports = ({ formService }) => (router) => {
           { ...pageData, ...curfewData, port },
           { filename, pdfOptions }
         )
+
         logger.info(`Returning rendered PDF for form '${templateName}'`)
         return pdf
       } catch (e) {
         logger.warn(`Caught an exception while rendering form ${templateName}: ${e}`)
-        return null
+        throw e
```
bell-pa, line 93: This stops the app hanging.  That's good.
It doesn't make the PDF renderer work though. Presumably that particular problem has gone away...

bell-pa (approved):
LGTM

PR #1255, opened 2021-01-11
DCS-775 Dropping external dependency on pdf library

andrewrlee:
This will allow us to more easily manage it's dependencies


bell-pa (commented):
I don't see any tests. Is that deliberate?

bell-pa (approved):
LGTM

PR #1256, opened 2021-01-14
DCS-775 Adding option to disable shm in bid to solve OOM issues

andrewrlee:
As described here: https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md\#tips


petergphillips (approved):
LGTM

PR #1257, opened 2021-01-19
DCS-755: Bug fix for rendering PDFs.

bell-pa:
:bug: Configure the Docker image according to the latest puppeteer recommendations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,11 +20,11 @@ WORKDIR /app
 # Install latest chrome dev package libs so that the bundled version of Chromium installed by Puppeteer will work
 # https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
 RUN apt-get update \
-    && apt-get install -y wget gnupg gnupg1 gnupg2 make python \
+    && apt-get install -y wget gnupg \
     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
     && apt-get update \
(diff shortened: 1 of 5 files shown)


petergphillips (approved):
LGTM

Dockerfile
```diff
@@ -20,11 +20,11 @@ WORKDIR /app
 # Install latest chrome dev package libs so that the bundled version of Chromium installed by Puppeteer will work
 # https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
 RUN apt-get update \
-    && apt-get install -y wget gnupg gnupg1 gnupg2 make python \
```
andrewrlee, line 23: slightly concerned as thought python was a dependency of the logging framework but I assume it's been tested and I'm probably wrong

package.json
```diff
@@ -8,7 +8,7 @@
     "npm": "^6.4.1"
   },
   "scripts": {
-    "copy-views": "cp -R server/views/ dist/server/views/",
+    "copy-views": "cp -R server/views/ dist/server/views/ && cp -R mock-server/routes/images dist/mock-server/routes",
```
andrewrlee, line 11: meh.. this seems a bit messy

Dockerfile
bell-pa: I'll check the logging.  Do you know why it is needed? Is there something I can read?

package.json
bell-pa: I know.  
mock-server needs to serve this image. Without it you can't print a licence against the mocks.
Its there so that you can run mock-server from a licences docker image using docker-compose.
It is one small image.
I could add a separate script and add that to the sequence called from 'build'.  Or have a separate Docker file for stubs. (More complicated but separate).  Any suggestions?

andrewrlee (approved):
LGTM

package.json
andrewrlee: nah that's alright...

PR #1259, opened 2021-02-01
DCS-824: updated readme 

GurnankCheema:
Updated readme with clearer instructions for running integration tests. 
The primary reason for this ticket was to verify whether a user is logged out when a second user logs in. 
Have manually produced the steps in DEV environment to confirm that this is the case.


andrewrlee (approved):
LGTM

PR #1260, opened 2021-02-02
DCS-772 Only show 'Exit to DPS' link to NOMIS users.

bell-pa:
Only show the 'Exit to DPS' links if the user has the 'PRISON' role.


server/authentication/roles.js
```diff
@@ -15,4 +15,5 @@ module.exports = {
   isAuthServiceRole: (role) => authServiceRoles.includes(role),
   isApplicationRole: (role) => applicationRoles.includes(role),
   isAdminRole: (role) => adminRole === role,
+  hasNomisAuthSource: (user) => user && user.authSource === 'nomis',
```
andrewrlee, line 18: think a role of PRISON is automatically added for all nomis users

andrewrlee (approved):
LGTM

server/data/nomisClientBuilder.ts
```diff
@@ -191,7 +191,7 @@ export = (token) => {
       return oauthRestClient.getResource(`/api/user/me`)
     },
 
-    getUserRoles(): Promise<Role> {
+    getUserRoles(): Promise<[Role]> {
```
andrewrlee, line 194: this is weird, is this a valid way to define array tyoes?
rather than `Role[]` or `Array<Role>`

server/routes/admin/admin.js
```diff
@@ -5,7 +5,10 @@ module.exports = () => (router) => {
   router.use(authorisationMiddleware)
 
   router.get('/', (req, res) => {
-    return res.render('admin/index', { exitUrl: links.exitUrl })
+    return res.render('admin/index', {
+      exitUrl: links.exitUrl,
+      showExitUrl: req.user && req.user.isPrisonUser,
```
andrewrlee, line 10: meh.. would assume you can do `req.user?.isPrisonUser` (but have never tried it)

andrewrlee (approved):
LGTM

PR #1261, opened 2021-02-04
DCS-772 Hide 'Exit to DPS link' from users for whom the link makes no sense: Bug fix.

bell-pa:
* Fix broken role handling in user.js. 
* Tweak Redis config so that it is possible to connect to 'dev' environment when running app locally. 
  * For running locally against 'dev' this means port-forwarding RDS and Redis ports to localhost, connecting using TLS using a local CA cert while ignoring mismatching domains. 
  * In production still use TLS and the cert, but enforce domain name matching.
  * When running locally against docker images disable TLS.

### DIFF
--- a/helm_deploy/licences/templates/_envs.tpl
+++ b/helm_deploy/licences/templates/_envs.tpl
@@ -29,7 +29,6 @@ env:
         key: database_password
 
   - name: DB_SSL_ENABLED
-    value: {{ .Values.env.DB_SSL_ENABLED }}
     value: "true"
 
   - name: APPINSIGHTS_INSTRUMENTATIONKEY
@@ -156,6 +155,5 @@ env:
         key: auth_token
 
   - name: REDIS_TLS_ENABLED
-    value: {{ .Values.env.REDIS_TLS_ENABLED }}
     value: "true"
 {{- end -}}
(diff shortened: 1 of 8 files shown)


andrewrlee (dismissed):
LGTM, but test failures

helm_deploy/licences/templates/_envs.tpl
```diff
@@ -28,6 +28,10 @@ env:
         name: dps-rds-instance-output
         key: database_password
 
+  - name: DB_SSL_ENABLED
+    value: {{ .Values.env.DB_SSL_ENABLED }}
+    value: "true"
```
andrewrlee, line 33: don't understand this but might have seen it before - is this a helm specific thing? 

would have thought as it produces yaml the first value would always be overriden with true?

bell-pa: No idea.  
Just copied what had been used for Redis since it seemed about right.
I assumed that the first 'value' fell through to the second if it wasn't present.  
That was just a wild guess based on what someone else had written.
Think I'd better find out what it really does - if anything!

bell-pa: I simplified this to 
```

andrewrlee (approved):
LGTM

PR #1264, opened 2021-02-18
DCS-851 Include YEAR please into the Timestamp on all events in Licences

bell-pa:
Date-time format is now DD/MM/YYYY HH:mm:ss for admin licence audit view.


GurnankCheema (approved):
LGTM

PR #1270, opened 2021-03-09
DCS-890 Use version 6 of puppeteer to align with chrome 89 installed in Docker container.

bell-pa:
The Docker build currently installs version 89.0.4839 of chrome to /opt/google/chrome 
According to the puppeteer documentation the only version of puppeteer that is compatible with chrome 89 is v6.0.0.
This combination appears to work correctly when the docker image is built and tested locally using the docker-compose set up in docker-compose-full.yml: PDF documents can be rendered repeatedly; chrome processes are terminated and removed after rendering is complete.

### DIFF
--- a/docker-compose-full.yml
+++ b/docker-compose-full.yml
@@ -7,8 +7,6 @@
 version: '3.8'
 services:
   licences:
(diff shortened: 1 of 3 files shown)


andrewrlee (approved):
LGTM

PR #1271, opened 2021-03-10
DCS-890 Fix high severity security advisory GHSA-p493-635q-r6gr

bell-pa:
New build to update to version 3.0.2 of pug to remove a high severity security advisory.

This service uses the Node.js package 'Puppeteer' to render PDF documents. Puppeteer is a Node library which provides a high-level API to control Chrome or Chromium. When used in a Docker image based on Debian Linux Puppeteer requires a compatible version of Chrome to be installed. Puppeteer either uses this or optionally installs its own compatible version of Chromium.  The bundled version of Chromium depends upon the libraries provided by the Google Chrome installation.
The last time this service was built and deployed to production the stable version of Google Chrome was 88. Now it is 89 and version 88 has been withdrawn.

I cannot find a combination of chrome 89 and puppeteer (either with or without its bundled version of chromium) that works for rendering PDF documents when run in the Kubernetes cluster. Most combinations don't work when run locally as a Docker image either.

Fortunately the package repos for apt contain version 88 of Chromium and the following Chromium based combination works, both as a Docker image run locally using docker-compose and when run in the cluster:
* Pug 3.0.2
* Puppeteer 7.1.0
* Debian package chromium=88.0.4324.182-1~deb10u1

It looks as if the Debian repositories only contain package versions corresponding to the current stable Chrome release and the release prior to that.  When Chromium version 88 is removed from the package repositories it will not be possible to build the application using this combination and the problem will recur.  Hopefully when that happens there will be a newer workin g combination of Puppeteer and Chromium or Google Chrome that the service can use in the cluster.


andrewrlee (approved):
LGTM

PR #1282, opened 2021-04-21
DCS-951 Replace Puppeteer with Gotenberg

bell-pa:
Some javascript unit tests are now integration tests. I moved them to the integration_test folder.  The tests need Gotenberg to be running - `docker-compose up ` does this.  The circle ci build now has an extra step to run these tests.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14.15-buster as builder
```
bell-pa, line 1: This Docker file is a pain.  It seems that the version of docker used by circleci build the docker image doesn't work very well with multistage builds. In theory this build can use node:14.16-buster-slim as 'base'. do `RUN apt-get update && apt-get upgrade -y` and then use this image for building the app (FROM base as builder) and also for creating the final release image (FROM base as release). That would be faster and simpler. Unfortunately this breaks the build at the mkdir /app step.  Clearly the builder and release images aren't separate.
Hence the version below. meh.

docker-compose-full.yml
```diff
@@ -8,6 +8,7 @@ version: '3.8'
 services:
```
bell-pa, line 8: docker-compose-full.yaml runs everything in docker.  You can run the groovy integration tests against this, or just point your browser at localhost:3000

helm_deploy/licences/Chart.yaml
```diff
@@ -1,5 +1,9 @@
-apiVersion: v1
```
bell-pa, line 1: Now has a nested chart for Gotenberg

helm_deploy/licences/charts/gotenberg/templates/_helpers.tpl
```diff
@@ -0,0 +1,61 @@
+{{/* vim: set filetype=mustache: */}}
```
bell-pa, line 1: Shamelessly copied form the parent chart

helm_deploy/licences/charts/gotenberg/values.yaml
```diff
@@ -0,0 +1,9 @@
+# This is a YAML-formatted file.
```
bell-pa, line 1: The values entered here are defaults.  They're all overridden in values-dev.yaml, values-preprod.yaml and values-prod.yaml

helm_deploy/licences/templates/_envs.tpl
```diff
@@ -153,4 +153,10 @@ env:
 
   - name: REDIS_TLS_ENABLED
     value: "true"
+
+  - name: HDC_URL
```
bell-pa, line 157: The HTML templates for forms and licences include links to CSS files served from within the public folded of licences.  HDC_URL is the prefix for those links.

helm_deploy/licences/templates/_envs.tpl
```diff
@@ -153,4 +153,10 @@ env:
...
+    value: "http://licences:80"
+
+  - name: GOTENBERG_API_URL
+    value: "http://licences-gotenberg:3000"
```
bell-pa, line 161: The URL prefix the licences app to use when POSTing HTML documents to the licences-gotenberg service for conversion to PDF.

server/config.js
```diff
@@ -124,14 +143,10 @@ module.exports = {
...
-          bottom: '70px',
-          left: '50px',
-          right: '30px',
-        },
-        displayHeaderFooter: true,
+        marginTop: '0.8',
+        marginBottom: '0.7',
+        marginLeft: '0.55',
+        marginRight: '0.35',
       },
```
bell-pa, line 150: Tweaked to be Gotenberg friendly.  A4 is default paper size so not needed.  Margins are in inches.

.circleci/config.yml
```diff
@@ -5,6 +5,15 @@ orbs:
...
+    docker:
+      - image: cimg/node:14.15-browsers
+      - image: thecodingmachine/gotenberg:6
+        user: gotenberg
+        environment:
+          DEFAULT_LISTEN_PORT: 3001
+    working_directory: ~/app
+
   integration-tests:
     docker:
```
bell-pa, line 18: The groovy integration tests don't test PDF generation, so a running Gotenberg image isn't required.

Dockerfile
```diff
@@ -1,4 +1,4 @@
...
+FROM node:14.16-buster as builder
```
andrewrlee, line 1: think if we bump this we should also bump the circle image

helm_deploy/licences/charts/gotenberg/templates/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: {{ .Values.image.port }}
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /ping
```
andrewrlee, line 49: does it have a ping?

bell-pa: Yes

andrewrlee (approved):
LGTM

PR #1288, opened 2021-06-02
DCS-925: Agency Notification and Licence Variation forms downladable …

GurnankCheema:
…as Word docs

Previosuly, just like all the other forms, these forms were presented in pdf format.
However it is not clear whether the Licences teams want these word versions after all.
So have made the changes to provide word docs but have also retained the code for the pdf versions.
Will not request a PR until the Licences teams are sure of  what they want.
Ticket put back into the backlog in the meantime.


GurnankCheema:
closed as too old

PR #1297, opened 2021-08-11
DCS-896 change of style and text of link leading back to DPS.

GurnankCheema:
This change applies to the Case list page only


louisenorris (approved):
LGTM

PR #1298, opened 2021-08-19
DCS-1160 Caseload actions added to each page

GurnankCheema:
User can view and change their location, aka caseload, from any page in the app.
This action was previously done by user via the DPS service.
Showing the 'Change location' link is conditional on user having more than 1 location.
Showing the current location is dependant on user having a currentlyActive caseload

### DIFF
--- a/server/services/userService.js
+++ b/server/services/userService.js
@@ -14,8 +14,8 @@ module.exports = (nomisClientBuilder) => {
 
     logger.info(`User profile success - username: ${username}`)
 
-    const activeCaseLoads = await nomisClient.getUserCaseLoads()
-    const activeCaseLoad = activeCaseLoads.find((cl) => cl.currentlyActive)
+    const caseLoads = await nomisClient.getUserCaseLoads()
+    const activeCaseLoad = caseLoads.find((cl) => cl.currentlyActive)
     const activeCaseLoadId = getIn(activeCaseLoad, ['caseLoadId'])
 
     return {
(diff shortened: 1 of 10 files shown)


server/services/userService.js
```diff
@@ -25,6 +25,7 @@ module.exports = (nomisClientBuilder) => {
       isPrisonUser,
       activeCaseLoad,
       activeCaseLoadId,
+      activeCaseLoads,
```
andrewrlee, line 28: can we just call this caseloads? 

`user.caseLoads` provides enough context - and will make the code easier to read in the pug template

server/views/layout.pug
```diff
@@ -71,6 +71,17 @@ html(lang="en")
 
     main.app-container(role="main")
       include ./includes/notice
+
+      div(class="change-location")
+        if user.activeCaseLoad && !user.activeCaseLoad.description.includes('---')
+          span(class="change-location__location" data-qa="active-location") #{user.activeCaseLoad.description}
+
+        if user.activeCaseLoads && user.activeCaseLoads.length > 1
```
andrewrlee, line 79: rather than define the same checks multiple times it would probably read better to extract out into readable variable names

`if hasMultipleCaseloads`
`if hasActiveCaseload`

`if hasMultipleCaseloads || hasActiveCaseload`

there should be other examples of how to do this in licences

GurnankCheema: goos idea

server/services/userService.js
GurnankCheema: yes

andrewrlee (approved):
LGTM

PR #1299, opened 2021-08-19
DCS-894 updated header

louisenorris:


server/services/userService.js
```diff
@@ -25,6 +26,7 @@ module.exports = (nomisClientBuilder) => {
       isPrisonUser,
       activeCaseLoad,
       activeCaseLoadId,
+      displayNameInitial: `${forenameToInitial(profile.name)}`,
```
louisenorris, line 29: Just to check - do we know that a user will always have a name in profile for us to use in forenameToInitial?

andrewrlee: should do

andrewrlee (approved):
LGTM

server/app.js
```diff
@@ -208,12 +208,22 @@ module.exports = function createApp({
...
       hostname: req.hostname,
+      authUrl: config.nomis.authUrl,
+      apiClientId: config.nomis.apiClientId,
     }
     next()
   }
 
   app.use(addTemplateVariables)
 
+  function addreturnURL(req, res, next) {
```
andrewrlee, line 219: super meh.. but `R` for `return`

louisenorris: pushing up a correction for this now - thought I might as well.

andrewrlee:
closing as superseded

PR #1306, opened 2021-09-07
DCS-1180: location link text changed

GurnankCheema:
The location link text, which is in every page underneath the header, changes to 'Change your roles or location' user has multiple roles.


GurnankCheema:
PS integ tests will fail until the new user (ca_ro_dm_user) is available in Auth service. It is currently under a PR

andrewrlee (approved):
LGTM

PR #1314, opened 2021-10-14
DCS-1166 Moving condition config over to typescript 

andrewrlee:
Also adding initial structure for a second version


JamieDraperUK (approved):
LGTM

PR #1319, opened 2021-11-11
DSC-1310 adding ingress annotations

michalnawrockidj:


helm_deploy/licences/values.yaml
```diff
@@ -3,3 +3,6 @@
 
 generic-prometheus-alerts:
   targetApplication: licences
+
+  ingress:
+    external-dns.alpha.kubernetes.io/aws-weight: "100"
```
andrewrlee, line 8: ingress needs to be top level and this needs to be in an annotations section

michalnawrockidj: fixed

andrewrlee (approved):
lgtm

PR #1320, opened 2021-11-15
DCS-1248 Adding ability to store conditions version against a licence

andrewrlee:


server/services/config/conditionsConfig.ts
```diff
@@ -11,384 +11,390 @@ type ConditionMetadata = {
   subgroup_name: string
 }
 
-const additionalConditions: Record<ConditionVersion, ConditionMetadata[]> = {
```
andrewrlee, line 14: Now a map as one to represent condition version as a number

GurnankCheema (approved):
LGTM

PR #1321, opened 2021-11-17
DCS-1248 Adding missing test cases and moving condition router over to typescript

andrewrlee:


server/routes/conditions.ts
```diff
@@ -1,11 +1,12 @@
...
+
 const { asyncMiddleware } = require('../utils/middleware')
 const createStandardRoutes = require('./routeWorkers/standard')
 const logger = require('../../log')
 const { getIn } = require('../utils/functionalHelpers')
 const formConfig = require('./config/licenceConditions')
 
-module.exports =
-  ({ licenceService, conditionsService }) =>
+export default ({ licenceService, conditionsService }: { licenceService: LicenceService; conditionsService: any }) =>
```
connormaglynn, line 9: Not 100% here but had a similar problem with changing module.exports to export default, linter seemed to want it though but think it might be causing the build error.

Didn't manage to find the solution as it was on the SOC stuff last week, but was going to have another nosey after I'd done the migration work 😅

connormaglynn (approved):
LGTM

server/routes/conditions.ts
andrewrlee: as discussed

PR #1323, opened 2021-11-18
DCS-1248 Persisting conditions 

andrewrlee:
Still need to consider the following: 
* When should conditions version be persisted if not previously set? (Its possible for a user to never select an optional question)
* Add migration
* Read value when populating list

### DIFF
--- a/server/data/licenceClient.ts
+++ b/server/data/licenceClient.ts
@@ -50,7 +50,7 @@ export class LicenceClient {
   // eslint-disable-next-line class-methods-use-this
   async getLicence(bookingId: number): Promise<CaseWithVaryVersion> {
     const query = {
-      text: `select licence, booking_id, stage, version, vary_version from licences where booking_id = $1`,
+      text: `select licence, booking_id, stage, version, vary_version, conditions_version from licences where booking_id = $1`,
       values: [bookingId],
     }
 
--- a/server/routes/conditions.ts
+++ b/server/routes/conditions.ts
@@ -1,4 +1,5 @@
 import { LicenceService } from '../services/licenceService'
+import { CURRENT_CONDITION_VERSION } from '../services/config/conditionsConfig'
 
 const { asyncMiddleware } = require('../utils/middleware')
 const createStandardRoutes = require('./routeWorkers/standard')
@@ -43,7 +44,7 @@ export default ({ licenceService, conditionsService }: { licenceService: Licence
       const { bookingId, action } = req.params
       const isChange = action === 'change'
 
-      licenceService.update({
+      await licenceService.update({
         bookingId,
         originalLicence: res.locals.licence,
         config: formConfig.standard,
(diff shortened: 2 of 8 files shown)


server/routes/conditions.ts
```diff
@@ -35,6 +36,41 @@ export default ({ licenceService, conditionsService }: { licenceService: Licence
       })
     }
 
+    router.post('/standard/:bookingId', audited, asyncMiddleware(postStandard))
```
andrewrlee, line 39: these were handled by the standard router but now we need additional bespoke code

server/routes/config/licenceConditions.js
```diff
@@ -8,21 +8,6 @@ module.exports = {
...
-          change: '/hdc/licenceConditions/additionalConditions/change/',
-        },
-        No: {
-          path: '/hdc/taskList/',
-          change: '/hdc/review/licenceDetails/',
-        },
-      },
-      path: '/hdc/taskList/',
-      change: '/hdc/review/licenceDetails/',
-    },
```
andrewrlee, line 25: no longer needed as part of the route

server/data/licenceClient.ts
```diff
@@ -50,7 +50,7 @@ export class LicenceClient {
   // eslint-disable-next-line class-methods-use-this
   async getLicence(bookingId: number): Promise<CaseWithVaryVersion> {
     const query = {
-      text: `select licence, booking_id, stage, version, vary_version from licences where booking_id = $1`,
+      text: `select licence, booking_id, stage, version, vary_version, conditions_version from licences where booking_id = $1`,
```
bell-pa, line 53: Has the conditions_version column already been added to the licences table?  Does it have a default value, PREVIOUS_CONDITION_VERSION perhaps, for pre-existing records?

server/routes/conditions.ts
```diff
@@ -35,6 +36,41 @@ export default ({ licenceService, conditionsService }: { licenceService: Licence
...
+        formName: 'standard',
+        postRelease: res.locals.postRelease,
+      })
+
+      const conditionsVersion = res.locals.licence.versionDetails?.conditions_version
+
+      if (!conditionsVersion) {
+        await licenceService.setConditionsVersion(Number(bookingId), CURRENT_CONDITION_VERSION)
+      }
+
```
bell-pa, line 62: Looks as if the conditions_version will be set to CURRENT_CONDITION_VERSION for every licence record seen by this handler.  Is that right?
Should this only happen if the record doesn't currently have an 'additional' section or if a new version of the licence is being created?  Not sure.

server/data/licenceClient.ts
andrewrlee: that's right

server/routes/conditions.ts
andrewrlee: I think we need a migration to set it if the licence has a `standard` form section. 

The section gets created whether or not an answer is selected.

So this would mean that we would set the condition to the default value if they've seen the conditions an clicked 
submit. 

I think we'll need another point where we set it though as users might not even click on this page.
The migration will need to respect that also

bell-pa (approved):
LGTM

PR #1327, opened 2021-12-07
DCS-1356 Handling missing user in header

andrewrlee:


server/views/layout.pug
```diff
@@ -71,20 +71,20 @@ html(lang="en")
 
     main.app-container(role="main")
       include ./includes/notice
+      if user
```
andrewrlee, line 74: just wrapped this entire block with an `if user`

GurnankCheema (approved):
LGTM

PR #1333, opened 2021-12-30
Moving condition service over to typescript class

andrewrlee:


server/services/conditionsService.ts
```diff
@@ -5,8 +5,14 @@ import * as builder from './licenceWithConditionsBuilder'
 import { getAdditionalConditionsConfig, standardConditions, CURRENT_CONDITION_VERSION } from './config/conditionsConfig'
 import { AdditionalConditions, Licence } from '../data/licenceTypes'
 
-export default function createConditionsService() {
-  function getFullTextForApprovedConditions(licence: Licence) {
+export default class ConditionsService {
```
andrewrlee, line 8: moving to class so get better type info than a factory method returning a dynamic object

test/routes/contact.test.ts
```diff
@@ -12,13 +12,13 @@ jest.mock('../../server/services/userAdminService')
...
 describe('/contact', () => {
-  let userAdminService: UserAdminService
-  let roService: RoService
+  let userAdminService: jest.Mocked<UserAdminService>
+  let roService: jest.Mocked<RoService>
   let signInService
 
   beforeEach(() => {
-    roService = new RoService(undefined, undefined)
-    ;(jest as any).mocked(roService).findResponsibleOfficer.mockResolvedValue({
```
andrewrlee, line 21: don't need to cast everywhere.

test/routes/vary.test.ts
```diff
@@ -1,4 +1,3 @@
-import { mocked } from 'ts-jest/utils'
```
andrewrlee, line 1: `ts-jest/utils` now spews out a load of warnings saying that its deprecated and to move over to a new version of jest. 

This just isn't needed though and it's generally cleaner to not used `mocked` at all. 

The rest of the changes in this PR are to remove usage of `mocked`

connormaglynn (approved):
lgtm 👍

GurnankCheema (approved):
LGTM

PR #1336, opened 2022-01-05
DCS-1166 Move condition routes over to use additional condition from licence

andrewrlee:


server/routes/conditions.ts
```diff
@@ -82,9 +80,9 @@ export default ({
       logger.debug('GET /additionalConditions')
 
       const { action, bookingId } = req.params
-      const licence = res.locals.licence?.licence
-      const bespokeConditions = licence?.licenceConditions?.bespoke || []
-      const conditions = conditionsServiceFactory.forVersion(CURRENT_CONDITION_VERSION).getAdditionalConditions(licence)
+      const record = res.locals.licence
+      const bespokeConditions = record.licence?.licenceConditions?.bespoke || []
+      const conditions = conditionsServiceFactory.forLicence(record).getAdditionalConditions(record.licence)
```
andrewrlee, line 85: rather than always using the current version, now read version from licence if possible. 
Only if that is not present then use the current version

server/services/licenceService.ts
```diff
@@ -570,7 +571,7 @@ export class LicenceService {
     return this.licenceClient.saveApprovedLicenceVersion(bookingId, template)
   }
 
-  setConditionsVersion(bookingId: number, conditionVersion: number) {
+  setConditionsVersion(bookingId: number, conditionVersion: ConditionVersion) {
```
andrewrlee, line 574: narrow these types so they are more specific

server/services/conditionsService.ts
```diff
@@ -2,13 +2,25 @@ import moment from 'moment'
...
 import { AdditionalConditions, Licence } from '../data/licenceTypes'
+import { LicenceRecord } from './licenceService'
+import { ConditionVersion } from '../data/licenceClientTypes'
 
 export class ConditionsServiceFactory {
   forVersion(version: ConditionVersion) {
     return new ConditionsService(version)
   }
+
+  forLicence(licence: LicenceRecord) {
```
connormaglynn, line 15: Maybe some unit tests around these two new methods?

connormaglynn (approved):
lgtm

PR #1337, opened 2022-01-05
DCS-1166 externalising condition version from licence conditions builder

andrewrlee:


server/services/conditionsService.ts
```diff
@@ -24,11 +24,15 @@ export class ConditionsServiceFactory {
 }
 
 export class ConditionsService {
-  constructor(private readonly version: ConditionVersion) {}
+  private readonly builder: LicenceWithConditionsBuilder
+
+  constructor(private readonly version: ConditionVersion) {
+    this.builder = new LicenceWithConditionsBuilder(version)
```
andrewrlee, line 30: move from a direct dependency to a local field which is parameterised with the same version of conditions to use as the main conditions service

server/services/licenceWithConditionsBuilder.ts
```diff
@@ -1,161 +1,166 @@
...
-  approvedOnly = false
-) {
-  const { additional, bespoke } = rawLicence.licenceConditions
-
-  const additionalConditionsJustification =
-    rawLicence?.licenceConditions?.conditionsSummary?.additionalConditionsJustification
-  const getObjectForAdditional = createAdditionalMethod(rawLicence, selectedConditionsConfig, inputErrors)
-
-  const populatedAdditional = Object.keys(additional)
-    .sort(orderForView(getAdditionalConditionsConfig(CURRENT_CONDITION_VERSION).map((condition) => condition.id)))
```
andrewrlee, line 54: The main gist of this PR is to remove it's direct reliance on `CURRENT_CONDITION_VERSION` as this will be read from the licence instead

louisenorris (approved):
LGTM

PR #1341, opened 2022-01-10
DCS-1166 Moving form resources over to read condition version from li…

andrewrlee:
…cence record


server/routes/forms.ts
```diff
@@ -15,20 +16,19 @@ const {
...
-        licence: { licence },
-        licenceStatus,
-      } = res.locals
+      const { prisoner, licence: licenceRecord, licenceStatus } = res.locals
 
+      const { licence } = licenceRecord
       const isBass = getIn(licenceStatus, ['decisions', 'bassReferralNeeded']) === true
       const isAp = getIn(licenceStatus, ['decisions', 'approvedPremisesRequired']) === true
 
+      const conditions = conditionsServiceFactory.forLicence(licenceRecord).getFullTextForApprovedConditions(licence)
```
andrewrlee, line 30: this has been pulled out of the service to the route as there didn't seem much need for the form service to interact with the condition service. 

It was easier to just deal with the full licence record here rather than to pass both through or to try and work out the version here with the condition factory and pass that through - to be read by the factory again at the lower level.

Also made adjusting the tests much more straightforward

server/services/formService.ts
```diff
@@ -146,8 +143,7 @@ export default class FormService {
     bookingId: number
     token: string
   }) {
-    const [conditions, prisoner, caMailboxes] = await Promise.all([
-      this.conditionsServiceFactory.forVersion(CURRENT_CONDITION_VERSION).getFullTextForApprovedConditions(licence),
```
andrewrlee, line 150: condition service isn't actually async, so this wasn't required

louisenorris (approved):
LGTM

PR #1342, opened 2022-01-12
DCS-1166 extracting out conditions to new location

andrewrlee:
Also removing the templates for some unused conditions


louisenorris (approved):
LGTM

PR #1349, opened 2022-01-28
DCS-1239 Altering form for attend_all to allow selecting more than one option

andrewrlee:


server/data/licenceTypes.ts
```diff
@@ -169,7 +169,7 @@ export type AdditionalConditionsV2 = {
   NO_COMMUNICATE_VICTIM?: any
   ATTENDDEPENDENCY?: any
   ATTEND_DEPENDENCY_IN_DRUGS_SECTION?: any
-  ATTEND_ALL?: any
+  ATTEND_ALL?: { appointmentProfessions: any }
```
andrewrlee, line 172: this whole typedef is just partial - needed to add this to be able to access it

server/services/config/conditions/v1/conditions.ts
```diff
@@ -14,8 +14,18 @@ enum GroupName {
   RESTRICTED = 'Restricted additional conditions',
 }
 
-export const getPersistedAbuseAndBehaviours = (conditions: AdditionalConditions) =>
-  (conditions as AdditionalConditionsV1)?.COMPLYREQUIREMENTS?.abuseAndBehaviours
+export const modifyAdditionalConditions = (someConditions: AdditionalConditions) => {
```
andrewrlee, line 17: before this used to just pass back the list of behaviours so that the licence builder could prefix a space to each of the items (barring the first)

As we now have the same requirement for a list of professions but only for v2, I've moved the modification logic into each version of the conditions so it can vary per version

server/services/config/conditions/v2/conditions.ts
```diff
@@ -21,8 +20,27 @@ enum GroupName {
...
+    Object.assign(
+      abuseAndBehavioursConditions,
+      abuseAndBehavioursConditions.map((condition, index) => (index > 0 ? ` ${condition}` : condition))
+    )
+  }
+
+  const appointmentProfessions = conditions?.ATTEND_ALL?.appointmentProfessions
+
+  if (Array.isArray(appointmentProfessions)) {
+    Object.assign(
```
andrewrlee, line 38: TBH - I don't really understand this, copied the existing behaviour - I've used it for objects but not arrays. seems to override the assignment of the existing variable with a new valu

server/services/config/conditions/v2/conditions.ts
```diff
@@ -47,13 +65,21 @@ export const conditions: ConditionMetadata[] = [
   {
     id: 'ATTEND_ALL',
     text: 'Attend all appointments arranged for you with [INSERT NAME], a [PSYCHIATRIST / PSYCHOLOGIST / MEDICAL PRACTITIONER] and co-operate fully with any care or treatment they recommend.',
-    user_input: 'appointmentName', // need to change?!
+    user_input: 'attendAll',
```
andrewrlee, line 68: had to create a new form as forms aren't versioned and appointmentName is still being used by v1

server/services/config/conditions/v2/conditions.ts
```diff
@@ -47,13 +65,21 @@ export const conditions: ConditionMetadata[] = [
...
     },
     group_name: GroupName.MAKING_OR_MAINTAINING_CONTACT,
     subgroup_name: null,
+    displayForEdit: (inputtedCondition) => {
+      const result = inputtedCondition.appointmentProfessions
+      const appointmentProfessions = typeof result === 'string' ? [result] : result
+      return {
+        ...inputtedCondition,
+        appointmentProfessions,
+      }
```
andrewrlee, line 81: need to handle what happens when a single value is passed back - its relatively easy to handle this in the html but didn't want to be inconsistent with the other forms

louisenorris (approved):
LGTM

PR #1356, opened 2022-02-02
DCS-1384 migrate prod to live

GurnankCheema:
prod ingress moved from blue to green.
This project hasn't used the 'generic-service' to construct values files so have updated the values-prod.yaml itself rather removing the annotations from dev and pre-prod to then apply in a values.yaml singularly.


petergphillips (approved):
LGTM

PR #1363, opened 2022-02-07
DCS-1239 Fixing pdf generation

andrewrlee:
The browser instance within the gotenberg container relied on being able to access the pdf css from the running application. This was being blocked by the cross origin resource policy so the pdf was being generated without any styling


GurnankCheema (approved):
lgtm

PR #1372, opened 2022-04-22
Fixing cache break

andrewrlee:


Dockerfile
```diff
@@ -23,6 +23,10 @@ RUN npm ci --no-audit && \
 RUN npm prune --production
 
 FROM node:14.17-buster-slim
+
+ARG BUILD_NUMBER
+ARG GIT_REF
```
andrewrlee, line 28: this is a bit inconsistent compared with use of force and the template project - this only has two build stages. The production image is built from scratch and has its own upgrade command.

This was being cached as it relied on access to env vars that are only present globally. 

Would be good to tidy up this..

petergphillips (approved):
LGTM

PR #1381, opened 2022-06-08
DCS-1591 agency_notification is PDF from outset, i.e bypasses Gotenberg

GurnankCheema:


server/routes/forms.ts
```diff
@@ -66,6 +68,11 @@ export default (formService: FormService, conditionsServiceFactory: ConditionsSe
         throw new Error(`unknown form template: ${templateName}`)
       }
 
+      if (pdfFormatForms.includes(templateName)) {
+        logger.info(`Returning pdf document for the '${templateName}' form`)
+        return res.sendFile(path.join(__dirname, `../views/forms/downloads/${templateName}.pdf`))
+      }
+
```
GurnankCheema, line 75: agency_notification.pdf is kept separate in the downloads folder rather than with the other pug files because need to be able to target it individually as it is already a pdf and thus not to be processed by Gotenberg

PR #1387, opened 2022-06-22
DCS-1637: ✨ Add user to request in App insights

connormaglynn:


server/applicationVersion.ts
```diff
@@ -0,0 +1,9 @@
+// eslint-disable import/no-unresolved,global-require
```
connormaglynn, line 1: moved these functions out to align with [how we do it in the template project](https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/applicationVersion.ts)

server/utils/azureAppInsights.ts
```diff
@@ -0,0 +1,57 @@
+import { config } from 'dotenv'
```
connormaglynn, line 1: Converted to typescript and aligned with how the [template project does things](https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/utils/azureAppInsights.ts) for consistency.

It was doing all the same stuff, just in a slightly different way

server/utils/azureAppInsights.ts
```diff
@@ -0,0 +1,57 @@
...
+    defaultClient.addTelemetryProcessor(addUserDataToRequests)
+    return defaultClient
+  }
+  return null
+}
+
+export function addUserDataToRequests(envelope: EnvelopeTelemetry, contextObjects: ContextObject) {
+  const isRequest = envelope.data.baseType === Contracts.TelemetryTypeString.Request
+  if (isRequest) {
+    const { username, activeCaseLoadId } = contextObjects?.['http.ServerRequest']?.res?.locals?.user || {}
```
connormaglynn, line 46: Added in the changes to add the user information. Not sure if want to log any other info? Whether they're a CA, RO, DM for example? 🤔

test/utils/azureAppInsights.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    })
+
+    it('handles missing user details', () => {
+      const envelope = createEnvelope({ other: 'things' })
+
+      addUserDataToRequests(envelope, {
+        'http.ServerRequest': {},
+      } as ContextObject)
+
+      expect(envelope.data.baseData.properties).toStrictEqual({
```
connormaglynn, line 69: checking that no other properties are added if they're undefined

server/index.js
```diff
@@ -1,5 +1,6 @@
+const { appInsights } = require('./utils/azureAppInsights')
```
connormaglynn, line 1: Moved this to the top, I know they usually say to import it as early as possible 

https://www.npmjs.com/package/applicationinsights
`Important: applicationinsights must be setup and started before you import anything else. There may be resulting telemetry loss if other libraries are imported first.`

server/utils/azureAppInsights.ts
```diff
@@ -0,0 +1,57 @@
...
+    envelope.data.baseData.properties = Object.assign(properties || {}, {
+      ...(username && { username }),
+      ...(activeCaseLoadId && { activeCaseLoadId }),
+    })
+  }
+  return true
+}
+
+initialiseAppInsights()
+export const appInsights = buildAppInsightsClient()
```
connormaglynn, line 57: Had to export this since it's passed to the eventPusher object - kept the naming of `appInsights`. Would have preferred defaultClient but that would have clashed - maybe telemetryClient? 🤔

louisenorris: Think it probably makes sense to leave it as `appInsights` since it's created from the `buildAppInsightsClient()` function - seems consistent, what about `appInsightsClient`? 🤔

louisenorris: Prob doesn't matter too much though..

louisenorris (approved):
LGTM

andrewrlee (approved):
LGTM

PR #1390, opened 2022-08-09
DCS-1767 add new extremism additional condition

louisenorris:
Additional extremism licence added, appears in additional licences list and if selected also shown under licence conditions section on generated pdf:
Screenshot 2022-08-10 at 12 12 13

Screenshot 2022-08-10 at 12 13 13


davidthomas22 (approved):
LGTM

PR #1391, opened 2022-08-25
DCS-1766: 🆙 upgrade Ingress apiVersion to networking.k8s.io/v1 and nginx-ingress-controller to v1.2 

connormaglynn:
This PR is intended to test downtime moving to the new Ingress apiVersion and nginx-ingress-controller.

It will be tested in dev and potentially pre-prod. 

If the downtime is acceptable. It will then be pushed through to production at a reasonable time.

If not, we will revert this PR and raise another in favour of the [ 0 downtime process](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/networking/Switch-ingress-to-v1-ingress-controller.html#switch-to-the-new-ingress-controller-with-zero-downtime).


andrewrlee (approved):
LGTM

PR #1395, opened 2022-11-09
DCS-1893 converted risk management files to typescript

louisenorris:


server/routes/risk.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export default ({ licenceService }: { licenceService: LicenceService }) =>
+  (router, audited) => {
+    const standard = createStandardRoutes({ formConfig, licenceService, sectionName })
+
+    router.get('/:formName/:bookingId', asyncMiddleware(getStandard))
+
+    function getStandard(req, res: Response<any, LicenceLocals>) {
+      const { formName, bookingId, action } = req.params
+      const { licenceSection, nextPath, pageDataMap } = formConfig[formName]
```
andrewrlee, line 19: we can hard code formName to riskManagement everywhere

server/routes/risk.ts
```diff
@@ -0,0 +1,56 @@
...
+      const dataPath = pageDataMap || ['licence', sectionName, licenceSection]
+      const riskVersion = licenceService.getRiskVersion(res.locals.licence.licence)
+
+      const rawData = getIn(res.locals.licence, dataPath) || {}
```
andrewrlee, line 23: we can hardcode this rather than use the `getIn` helper

```
res.locals.licence?.licence?.risk?.riskManagement || {}
```

server/routes/risk.ts
```diff
@@ -0,0 +1,56 @@
...
+      const data =
+        firstItem(req.flash('userInput')) || licenceService.addSplitDateFields(rawData, formConfig[formName].fields)
```
andrewrlee, line 25: If there are no dates we could presumably remove this `licenceService.addSplitDateFields(rawData, formConfig[formName].fields)` (though I'm not familiar with this)

If we do need to keep it then we can hardcode the formName

server/routes/risk.ts
```diff
@@ -0,0 +1,56 @@
...
+      const errorObject = firstItem(req.flash('errors')) || {}
+
+      const viewData = { bookingId, data, nextPath, errorObject, action, sectionName, formName, version: riskVersion }
```
andrewrlee, line 28: we can hardcodee sectionName/formName though I'm surprised they are used.. We maybe able to reduce this viewData to only that which is required by this screen

server/routes/risk.ts
```diff
@@ -0,0 +1,56 @@
...
+        userInput: req.body,
+        licenceSection: 'risk',
+        formName: 'riskManagement',
+        postRelease: res.locals.postRelease,
+      })
+
+      return res.redirect(isChange ? `/hdc/review/risk/${bookingId}` : `/hdc/taskList/${bookingId}`)
+    }
+
+    router.get('/:formName/:action/:bookingId', asyncMiddleware(standard.get))
```
andrewrlee, line 52: we should be able to hardcode the `formName` if we remove reliance on this on line 18

GurnankCheema (approved):
LGTM

davidthomas22 (approved):
LGTM

PR #1398, opened 2022-11-29
DCS-1907: 🆙 Update Dependencies

connormaglynn:
**Updates:**
- npm dependencies 
	- only minor/patch versions of production dependencies
	- for test deps I updated everything bar linting and node types
-  hmpps orb to v6
	- We've updated from v3 in other projects, only required change is to remove the docker version which we don't have
- 	Gradle Wrapper update
-  Add a MockTestUi run task so we don't have to manually change the config to run with a UI


andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #1399, opened 2022-11-30
DCS-1907: ✨ Add Standard Conditions Versioning For Licences

connormaglynn:
**Tasks**
- [x] ✨ Add Standard Condition via DB Migration
- [x] ✨ Add functionality to store/retreive standard licence version in `licenceClient.ts`
- [x] ✨ Add functionality to store/retreive standard licence version in `licenceService.ts`
- [x] ♻️ Refactor `ConditionsVersion` to more specific `AdditionalConditionVersion` throughout the application

**Notes**
- Most of the file changes are due to the renaming `AdditionalConditionVersion` - this has been done to clarify the difference between itself and `StandardConditionsVersion`
- This change is not user-facing; this PR only sets up the ability to retrieve and persist the `StandardConditionsVersion` - a different ticket will implement this in the user-facing journeys

### DIFF
--- a/server/data/licenceClientTypes.ts
+++ b/server/data/licenceClientTypes.ts
@@ -2,6 +2,8 @@ import { Licence, LicenceStage } from './licenceTypes'
 
 export type ConditionVersion = 1 | 2
 
+export type StandardConditionsVersion = 1
+
 export type ConditionMetadata = {
   id: string
   text: string
@@ -20,6 +22,7 @@ export interface Case {
   stage: LicenceStage
   version: number
   additional_conditions_version: ConditionVersion
+  standard_conditions_version: StandardConditionsVersion
 }
 
 export interface CaseWithApprovedVersion extends Case {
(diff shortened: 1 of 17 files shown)


server/data/licenceClientTypes.ts
```diff
@@ -1,6 +1,8 @@
 import { Licence, LicenceStage } from './licenceTypes'
 
-export type ConditionVersion = 1 | 2
+export type AdditionalConditionsVersion = 1 | 2
+
+export type StandardConditionsVersion = 1
```
connormaglynn, line 5: Set this as just 1 for now. We'll add version 2 in the next ticket to implement the usage of these. Not entirely sure yet if the default/historical will be 1 or null.

andrewrlee: some historical will require a migration once we know how to identify those

andrewrlee (approved):
LGTM

PR #1400, opened 2022-12-01
DCS-1908: ✨Add v2 of `StandardConditions`

connormaglynn:
**Tasks**

- [x] ✨Add v2 of StandardConditions
- [x] ♻️ Refactor additionalConditions into their own sub folder

**Notes**
- Changes the file structure for `conditions` to `additionalConditions` and adds the new `standardConditions/v2` structure


server/data/licenceClient.ts
```diff
@@ -86,11 +86,12 @@ export class LicenceClient {
...
     stage: string = LicenceStage.DEFAULT,
     version: number = 1,
-    varyVersion: number = 0
+    varyVersion: number = 0,
+    standardConditionsVersion: StandardConditionsVersion = undefined
   ): Promise<number> {
     const query = {
-      text: 'insert into licences (booking_id, licence, stage, version, vary_version) values ($1, $2, $3, $4, $5)',
-      values: [bookingId, licence, stage, version, varyVersion],
+      text: 'insert into licences (booking_id, licence, stage, version, vary_version, standard_conditions_version) values ($1, $2, $3, $4, $5, $6)',
```
connormaglynn, line 93: I wasn't thrilled about having to amend some production code to create integration tests. The `utils` resource should use a separate client and not be coupled to the production implementation, in my opinion.

Happy to refactor the utils into its own client as part of this change. It'll just take extra time.

server/services/conditionsService.ts
```diff
@@ -75,8 +76,10 @@ export class ConditionsService {
     }
   }
 
-  getStandardConditions() {
-    return standardConditions
+  getStandardConditionsByVersion(standardConditionsVersion: StandardConditionsVersion) {
+    if (standardConditionsVersion == 1) return standardConditionsV1
```
connormaglynn, line 80: This line is technically redundant due to the default being V1, but thought the clarity of it was more important

server/services/conditionsService.ts
```diff
@@ -45,7 +46,7 @@ export class ConditionsService {
 
   // form generation
   getFullTextForApprovedConditions(licence: Licence) {
-    const standardConditionsText = standardConditions.map((it) => it.text.replace(/\.+$/, ''))
+    const standardConditionsText = standardConditionsV1.map((it) => it.text.replace(/\.+$/, ''))
```
connormaglynn, line 49: Bit of blurred lines here between the tickets. I thought I'd not jump down the rabbit hole of updating this line to dynamically get the versions, as it does feel like there will be a fair few changes needed to do it consistently with the current approach. Plus, the other changes implemented are user-facing, but this presumably does open the door to visual inconsistency.

That being said, I could add a parameter to the function for the version and use the `getStandardConditionsByVersion` method.

Happy either way 😄

server/services/config/conditions/standardConditions/v2/standardConditions.ts
```diff
@@ -0,0 +1,31 @@
...
+    text: 'Reside permanently at an address approved by the supervising officer and obtain the prior permission of the supervising officer for any stay of one or more nights at a different address.',
+  },
+  {
+    text: 'Not undertake work, or a particular type of work, unless it is approved by the supervising officer and notify the supervising officer in advance of any proposal to undertake work or a particular type of work.',
+  },
+  {
+    text: 'Not travel outside the United Kingdom, the Channel Islands or the Isle of Man except with the prior permission of your supervising officer or for the purposes of immigration deportation or removal.',
+  },
+  {
+    text: 'Tell your supervising officer if you use a name which is different to the name or names which appear on your licence.',
```
connormaglynn, line 26: Not entirely sure where these should of gone, so just stuck them at the bottom. Happy to move them around if we know what the order is?

server/services/conditionsService.ts
connormaglynn: Scratch what I said above; the int tests weren't going to work without updating it 😂 

I've now added the parameter in the class. I've added a unit test around this which is a bit narly with how it sits with other tests.

The following ticket will encompass refactoring this area, though, so we can better handle different versions of additional and standard conditions.

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/assessment/StandardConditionsVersionsSpec.groovy
```diff
@@ -0,0 +1,68 @@
...
+import spock.lang.Stepwise
+import uk.gov.justice.digital.hmpps.licences.pages.TaskListPage
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.LicenceConditionsAdditionalPage
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.LicenceConditionsStandardPage
+import uk.gov.justice.digital.hmpps.licences.pages.assessment.LicenceConditionsSummaryPage
+import uk.gov.justice.digital.hmpps.licences.util.Actions
+import uk.gov.justice.digital.hmpps.licences.util.TestData
+
+@Stepwise
+class StandardConditionsVersionsSpec extends GebReportingSpec {
```
connormaglynn, line 14: Happy for my Geb testing to be critqued, never done it before so I'm sure there'll be some mistakes here 😂

server/data/licenceClient.ts
connormaglynn: 👆 This is assuming the utils endpoint isn't a production endpoint - from what I could tell, it wasn't; I'm pretty sure we disable it in production with the env variables. I might be misremembering though 😅

docker-compose-minimal.yml
```diff
@@ -0,0 +1,27 @@
+version: '3.8'
```
andrewrlee, line 1: this doesn't seem related to the story and isn't reference anywhere?

connormaglynn: I use this to spin up a local development environment with minimal local docker images (postgres and redis). Happy to put a line in the readme to reference it 😄

andrewrlee: if you think its worth adding then it's probably worth adding to the README

andrewrlee (dismissed):
LGTM

server/services/config/conditionsConfig.ts
```diff
@@ -1,10 +1,10 @@
 import { AdditionalConditionsVersion, ConditionMetadata } from '../../data/licenceClientTypes'
 import { AdditionalConditions } from '../../data/licenceTypes'
-import * as v1 from './conditions/v1/conditions'
-import * as v2 from './conditions/v2/conditions'
-import { standardConditions as stdConditions } from './conditions/standardConditions'
+import * as v1 from './conditions/additionalConditions/v1/conditions'
+import * as v2 from './conditions/additionalConditions/v2/conditions'
```
andrewrlee, line 4: these should be  `additionalConditionsV1`, `additionalConditionsV2` ?

andrewrlee (approved):
LGTM

PR #1401, opened 2022-12-05
✨ Add functionality to display `StandardConditionsV2`

connormaglynn:


test/routes/licenceConditions.test.ts
```diff
@@ -21,7 +21,7 @@ describe('/hdc/licenceConditions', () => {
   let conditionsServiceFactory = createConditionsServiceFactoryStub()
 
   beforeEach(() => {
-    conditionsService.getStandardConditions.mockReturnValue([{ text: 'Not commit any offence' }])
+    conditionsService.getStandardConditionsByVersion.mockReturnValue([{ text: 'Not commit any offence' }])
```
andrewrlee, line 24: worth verifying what you are passing?

server/services/conditionsService.ts
```diff
@@ -75,17 +86,19 @@ export class ConditionsService {
     }
   }
 
-  getStandardConditions() {
+  getStandardConditionsByVersion(standardConditionsVersion: StandardConditionsVersion) {
+    if (standardConditionsVersion == 1) return standardConditionsV1
+    if (standardConditionsVersion == 2) return standardConditionsV2
```
andrewrlee, line 91: this logic should live in ConditionService like `getAdditionalConditionsConfig`.

This class shouldn't reference specific condition versions

server/services/conditionsService.ts
```diff
@@ -5,24 +5,30 @@ import {
...
   forLicence(licence: LicenceRecord) {
-    const version = licence.versionDetails?.additional_conditions_version || CURRENT_CONDITION_VERSION
-    return this.forVersion(version)
+    const additionalConditionsVersion =
+      licence.versionDetails?.additional_conditions_version || CURRENT_CONDITION_VERSION
+    const standardConditionsVersion = licence.versionDetails?.standard_conditions_version
+    return this.forVersion(additionalConditionsVersion, standardConditionsVersion)
   }
 
   getNewVersion(licence: LicenceRecord): AdditionalConditionsVersion {
```
andrewrlee, line 34: `getNewVersion`-> `getNewAdditionalConditionsVersion`

server/services/conditionsService.ts
andrewrlee: also might be good to make the defaulting explicit - only if there are no set of conditions present for a version, easy to miss if we add a version 3

PR #1404, opened 2022-12-08
DCS-1896 add new risk version content changes to pdf and CA read view…

louisenorris:
… tasklist


server/views/review/includes/licenceSummary.pug
```diff
@@ -71,10 +72,13 @@ div.largeMarginBottom
     div.borderBottom
       +titleRow('Risk', 'riskDetails')
         +roCase
-          +change("/hdc/risk/riskManagement/change/" + bookingId, 'riskEditLink')
+          +change("/hdc/risk/riskManagement/change/" + bookingId, 'riskEditLink')   
 
       +errorMessageOrBlock(errorObject.risk && errorObject.risk.riskManagement, 'risk-error')
-        include riskManagement
+        if riskVersion1
```
louisenorris, line 78: Check for version 1, as if not present on licence, risk management version 2 content should be displayed as the default here . 

The CA can view the licence summary before an RO answers risk management questions therefore it is possible that the version won't be present (version is persisted once an RO saves content in the risk management section for the first time) and if risk version is not present it will be a new licence and require the version 2 risk questions.

connormaglynn (approved):
LGTM 🚀

PR #1409, opened 2022-12-14
DCS-1958 migration script refactoring

michalnawrockidj:


migrations/20221208172816_add-riskManagement-init-value.js
```diff
@@ -1,13 +1,13 @@
 exports.up = async function up(knex) {
   await knex.schema.raw(`
-    update licences set licence = jsonb_set(licence,'{risk, riskManagement, version}','"1"')
-    where licence -> 'risk' -> 'riskManagement' -> 'version' is null
+    update licences set licence = licence #- '{risk, riskManagement, version}'
```
andrewrlee, line 3: isn't this removing it rather than setting it?

michalnawrockidj: good point I tested removal and committed by accident ;-)

andrewrlee (approved):
lgtm

PR #1412, opened 2022-12-15
DCS-1962 mandatory check banner fix

louisenorris:
Changing bass property conditional to check bassRequested instead of bassAreaSuitable.


andrewrlee (approved):
lgtm

PR #1413, opened 2022-12-16
DCS-1963 make risk view default content version 2

louisenorris:
To avoid old risk section questions displaying if no risk version present have changed conditionals to check for risk version 1 which following the data migration should always be present for old or in progress licences. Now version 2 content will be displayed as default.


davidthomas22 (commented):
LGTM

davidthomas22 (approved):

PR #1415, opened 2023-01-06
DCS-1984 removed unused content related to clearing office email

louisenorris:
Can I check the below code which is located in the `server/views/admin/licences/notify.pug` template... 

I think the _"and the clearing office"_ content just above the `form` is now the last remaining trace of the clearing office in the HDC service and that the submit button below no longer sends an email to the clearing office as well as the COM/RO functional mailbox. I think [this old commit](https://github.com/ministryofjustice/licences/commit/efb73342939c8ba0a279a725fbe46eca62544854) shows that functionality being removed from `server/services/notifications/roNotificationSender.js` but would value a second opinion on that. 

```
  h1.heading-large Notify Community Offender Manager
  include ../../taskList/prisonerDetails
  div.borderBottom.paddingBottom
    h2.heading-medium Notify
    div
      p Notify the COM that they have been sent the offender's case.
      p Pressing the button below will send an email to 
        a(href="/hdc/contact/" + bookingId) #{prisonerInfo.com.name}
        |, their functional mailbox and the clearing office.
      form(method='POST' action='/admin/licences/' + bookingId + '/notifyRo')
        input(type="hidden" name="_csrf" value=csrfToken)
        input.requiredButton.button(type="submit" value="Send emails")
```

Could add the removal of that reference to the clearing office above to this PR if that sounds right to others 😅


michalnawrockidj (approved):
LGTM

PR #1421, opened 2023-01-12
DCS-1974: ⬆️ update data-extractor to v0.1.4 🔧 Fix int_test pipeline failing

connormaglynn:


.circleci/config.yml
```diff
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@6
-  browser-tools: circleci/browser-tools@1.1.1
+  hmpps: ministryofjustice/hmpps@7
```
connormaglynn, line 4: The breaking change here is to reduce the timeout of deployments from 15minutes to 5. We usually take around 2 minutes to deploy. So no further changes are required.

helm_deploy/licences/Chart.yaml
```diff
@@ -12,5 +12,5 @@ dependencies:
     version: 0.4.1
     repository: https://ministryofjustice.github.io/hmpps-helm-charts
   - name: generic-data-analytics-extractor
-    version: 0.1.3
+    version: 0.1.4
```
connormaglynn, line 15: This now uses the python scripts. We're having some issues with the cronjob, so though updating to use the python scripts may help / be easier to get help with

louisenorris (approved):
LGTM

PR #1430, opened 2023-03-09
Update dockerfile

davidthomas22:


psoleckimoj (approved):
LGTM!
We can sit down and see if it's possible to sort your Dockerfile so it's nearer the template one at some point.

PR #1433, opened 2023-03-13
Update Dockerfile

psoleckimoj:


DockerfileNEW
```diff
@@ -0,0 +1,80 @@
...
+
+ARG BUILD_NUMBER=1_0_0
+ARG GIT_REF=not-available
+
+RUN npm install -g npm@9
+
+RUN apt-get install -y make python3 g++
+
+COPY package*.json ./
+RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit
```
davidthomas22, line 40: Gradle tests used for Integration tests, so I guess the cypress reference can be removed?

DockerfileNEW
```diff
@@ -0,0 +1,80 @@
+# Stage: base image
+FROM node:18.12-bullseye-slim as base
+
+ARG BUILD_NUMBER=1_0_0
```
davidthomas22, line 4: In current app this is defined as below. Is this ok?

`ARG BUILD_NUMBER
ARG GIT_REF`

davidthomas22 (commented):
Missing the 

`# Create app directory
RUN mkdir /app && \
    chown appuser:appgroup /app`
    
    is this required?

DockerfileNEW
```diff
@@ -0,0 +1,80 @@
+# Stage: base image
+FROM node:18.12-bullseye-slim as base
```
davidthomas22, line 2: Currently, uses builder, will the switch to base impact? I guess not due to it being refactored to build from base?

`FROM node:18.12-bullseye-slim as builder`

PR #1437, opened 2023-03-23
DCS-2041 fixing build

michalnawrockidj:


Dockerfile
```diff
@@ -46,6 +44,9 @@ RUN export BUILD_NUMBER=${BUILD_NUMBER} && \
 
 RUN npm prune --no-audit --production
 
+RUN apt-get autoremove -y && \
```
andrewrlee, line 47: no point doing it here as the base image will already have the dependencies in it so contribute to the final image size - you have to do it as part of the update/upgrade stage...

michalnawrockidj: Screenshot 2023-03-24 at 10 56 07
this is a current size of the image comparing to the last successful one

PR #1458, opened 2023-05-05
DCS-2209 added presumed unsuitable reasons

louisenorris:
Added `Category A prisoner` and `Sentenced to 4 years or more by an overseas court and repatriated to the UK with less than 4 years to serve` presumed unsuitable reasons:
Screenshot 2023-05-05 at 10 46 56


GurnankCheema (approved):
LGTM

PR #1462, opened 2023-05-05
DCS-2216 cancellation variation form content changes

louisenorris:


server/views/forms/cancel_agency_notification.pug
```diff
@@ -25,10 +25,10 @@ block content
     tr.spaceUnder
       td Curfew address: #{CURFEW_ADDRESS}
 
-  p It has been decided that the above offender should not be released on HDC; or the following details of the agency notification previously issued, are amended as below.
+  p It has been decided that the above offender should not be released on HDC; or the following details of the Agency Notification issued on &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; are amended as below.
```
GurnankCheema, line 28: Is there a better way of adding the spaces rather than &nbsp?

louisenorris: I haven't found anything as straightforward and licences uses this in the pug files elsewhere.. Could have a dotted line space maybe as for the user to write on. I'll check with Design to see what's preferred.

louisenorris: Design colleagues have confirmed a dotted line would be better so updating this shortly.

GurnankCheema (approved):
LGTM

PR #1474, opened 2023-05-24
DCS-2241 add new page to DM approval journey

GurnankCheema:
New page for a DM 'consideration' added

DM's Final decision task button will now go to new 'consideration' page instead of the refuse/allow page.
However if in previous questions, the answer to Mandatory Address checks is no, then the task list button will lead to the Mandatory Address Check warning page first (this is as per current behaviour)

**new page**

Screenshot 2023-05-24 at 12 57 57

**new label for Final decision task if user answers No**

Screenshot 2023-05-24 at 15 28 11

**and in caselist**

Screenshot 2023-05-24 at 15 31 09

### DIFF
--- a/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/decision/ConsiderationSpec.groovy
+++ b/licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/decision/ConsiderationSpec.groovy
@@ -1,0 +1,94 @@
+package uk.gov.justice.digital.hmpps.licences.specs.decision
+
+import geb.spock.GebReportingSpec
+import spock.lang.Shared
+import spock.lang.Stepwise
+import uk.gov.justice.digital.hmpps.licences.pages.TaskListPage
+import uk.gov.justice.digital.hmpps.licences.pages.decision.ApprovalConsiderationPage
+import uk.gov.justice.digital.hmpps.licences.pages.decision.ApprovalReleasePage
+import uk.gov.justice.digital.hmpps.licences.util.Actions
+import uk.gov.justice.digital.hmpps.licences.util.TestData
+
+@Stepwise
+class ConsiderationSpec extends GebReportingSpec {
+
+  @Shared
+  TestData testData = new TestData()
+  @Shared
+  Actions actions = new Actions()
+
+  def setupSpec() {
+    testData.loadLicence('decision/unstarted')
+    actions.logIn('DM')
+  }
+
+  def cleanupSpec() {
+    actions.logOut()
+  }
+
+  def 'Starts with nothing selected'() {
+
+    when: 'I view the consideration page'
+    to ApprovalConsiderationPage, testData.markAndrewsBookingId
+
+    then: 'Neither radio option is selected'
+    considerationRadios.checked == null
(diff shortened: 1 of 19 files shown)


server/services/licence/licenceStatusLabels.js
```diff
@@ -22,6 +22,10 @@ const status = {
   licenceCreated: { statusLabel: 'Licence created', activeCase: true },
   notComplete: { statusLabel: 'Not complete', activeCase: true },
   inProgress: { statusLabel: 'In progress', activeCase: true },
```
GurnankCheema, line 24: This will be the task list label if DM selects No at the new consideration page

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/decision/ConsiderationSpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+import spock.lang.Shared
+import spock.lang.Stepwise
+import uk.gov.justice.digital.hmpps.licences.pages.TaskListPage
+import uk.gov.justice.digital.hmpps.licences.pages.decision.ApprovalConsiderationPage
+import uk.gov.justice.digital.hmpps.licences.pages.decision.ApprovalReleasePage
+import uk.gov.justice.digital.hmpps.licences.util.Actions
+import uk.gov.justice.digital.hmpps.licences.util.TestData
+
+@Stepwise
+class ConsiderationSpec extends GebReportingSpec {
```
louisenorris, line 13: For consistency should this be `ApprovalConsiderationSpec`?

server/services/licence/licenceStatusTypes.ts
```diff
@@ -54,6 +54,7 @@ export type Decisions = {
   dmRefused?: boolean
   refusalReason?: string
   decisionComments?: string
+  dmNotConsidered?: boolean
```
louisenorris, line 57: I think the naming of `dmNotConsidered` is a bit confusing as if a user answers No they have not considered its true and Yes is false. Maybe makes more sense to call it `dmConsidered` so a truthy value is yes? I know we're flagging that something has not been done but in terms of setting out the logic in code think it's easier to read as `dmConsidered` 🤔

louisenorris: Hope I'm understanding correctly 😅 have confused myself

licences-specs/src/test/groovy/uk/gov/justice/digital/hmpps/licences/specs/decision/ConsiderationSpec.groovy
GurnankCheema: agreed

GurnankCheema: yes

server/services/licence/licenceStatusTypes.ts
louisenorris: Understand the nature of this now and it will stay as is 😄

louisenorris (approved):
LGTM

PR #1475, opened 2023-05-24
DCS-2245 update docker compose files

GurnankCheema:
The latest images that would be pulled down when doing a docker-compose pull will cause auth error when starting
the integration tests. This is because of a has been a change to the token-verification-api 

It now uses SPRING_DATA_REDIS_HOST rather than SPRING_REDIS_HOST


andrewrlee (approved):
LGTM

PR #1478, opened 2023-05-25
DCS-2237 version  the main question text in curfew address review page

GurnankCheema:
The text of the main question of the /curfew/curfewAddressReview/{bookingId} and the message displayed in /review/LicenceDetails/{bookingId} has changed. The old versions are to remain for old and partly started licences. THe new version to be used from 6/6/23 onwards
These are the pages the codes affects


server/routes/config/curfew.js
```diff
@@ -86,21 +86,23 @@ module.exports = {
...
         },
       },
+      {
+        consentHavingSpoken: {
+          validationMessage: 'Say if you were able to speak to the main occupier and if they consented to HDC',
+        },
+      },
       {
         electricity: {
-          dependentOn: 'consent',
```
GurnankCheema, line 91: the dependantOn and predicate are not required because the functionality is handled by the bespokeAddressSchema

server/views/curfew/approvedPremisesAddress.pug
```diff
@@ -1,5 +1,4 @@
 extends ../layout
-include ../proposedAddress/includes/addressDetail
```
louisenorris, line 2: Just a check but definitely don't need this?

GurnankCheema: I'm pretty sure as I renamed addressDetail to either addressDetailV1 or V2. Checked it wasn't used in other places too.

server/services/licence/curfewAddressState.ts
```diff
@@ -73,17 +73,17 @@ const approvedPremisesAddressState = (licence) => {
...
 
   if (offenderIsOccupier && electricity) {
     return TaskState.DONE
   }
-  if (consent && electricity) {
+  if ((consent && electricity) || (consentHavingSpoken && electricity)) {
     return TaskState.DONE
   }
-  if (consent || electricity) {
+  if (consent || consentHavingSpoken || electricity) {
```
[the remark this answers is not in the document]
GurnankCheema: So I didn't want to introduce another 'if' as I think that's more confusing.
What the change does is is extend the circumstances as to  when the task is 'done' 
So rather than just checking against consent or  electricity, it's now additionally checking for consentHavingSpoken
so  _if (consent || electricity)_  goes to   _if (consent || consentHavingSpoken || electricity)_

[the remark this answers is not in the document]
louisenorris: ok

server/views/curfew/approvedPremisesAddress.pug
louisenorris: And so do we not want the address details displaying here anymore?

GurnankCheema: The addressDetail component isn't currently being rendered in the application at the moment. Although it's been 'included' on line 2 of approvedPremisesAddress.pg it's not being called elsewhere hence I removed it

louisenorris: Great will approve

PR #1479, opened 2023-05-26
DCS-2240 risk mental healthcare plan question fix

louisenorris:
Small fix to make sure that prison healthcare not consulted warning only displayed when a user selects Yes that it is essential for a mental healthcare plan to be in place and selects No to consulted with prison healthcare.

Screenshot 2023-05-26 at 11 33 38


andrewrlee (approved):
LGTM

PR #1483, opened 2023-06-01
DCS-2218 versioning postpone pdf form

louisenorris:
Added versioning to postpone PDF form to display content depending on postpone version. 

**If a licence has a postpone version of 1 the PDF content will be:**
Screenshot 2023-06-01 at 22 07 35
**If a licence has a postpone version of 2 the PDF content will be:**
Screenshot 2023-06-01 at 22 04 00
**By default if no version is present yet (version only persisted when postpone section is completed) the PDF content will be as version 2 but not include the postpone reason:**
Screenshot 2023-06-01 at 22 03 32
**Note:** 
A migration updating all licences with a postpone section present with a postpone version of 1 is required. This is because if a postpone version of 1 doesn’t exist the service will show version 2 content in the PDF. A user can access the PDF forms prior to completing the relevant section therefore by default we want to show version 2 content and only version 1 if they have completed the postpone section prior to this versioning taking place.

### DIFF
--- a/integration_test/routes/forms.test.ts
+++ b/integration_test/routes/forms.test.ts
@@ -123,6 +123,65 @@ describe('/forms/', () => {
       expect(pdfText).toContain('Mark Andrews You are eligible for early release')
       expect(pdfText).toContain('you could be released from prison on 23rd August 2019')
     })
+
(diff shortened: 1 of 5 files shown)


andrewrlee (approved):
lgtm

PR #1484, opened 2023-06-02
DCS-2246 add migration file to add postpone version to active/started licences

louisenorris:
This will set a postpone version of '1' for all currently 'active' and 'already started but not completed' licences. Is this correct for to not add to completed licences in the case of a postpone version (checking as in the case of viewing the postponed pdf form if a postpone version of 1 doesn’t exist the service will show version 2 content in the PDF. A user can access the PDF forms prior to completing the relevant section therefore by default we want to show version 2 content and only version 1 if they have completed the postpone section prior to this versioning taking place. Therefore for completed licences version 2 of the postponed pdf will display if viewed but I think this is all good). 

Not sure if the timestamp at the start of the file has to be completely exact - hard to so with seconds 😅


andrewrlee (approved):
LGTM

PR #1489, opened 2023-07-19
CVSL-1209 Updating deps, alert channels and speeding up tests

andrewrlee:


.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           key: dependency-cache-{{ checksum "package-lock.json" }}
       - run:
           name: unit tests
-          command: npm run test
+          command: npm run test:ci
```
andrewrlee, line 104: this run the tests in line - slower than in development but still much faster than before

package.json
```diff
@@ -18,7 +18,8 @@
     "start:dev": "npm run build && concurrently -k -p \"[{name}]\" -n \"Views,TypeScript,Node\" -c \"yellow.bold,cyan.bold,green.bold\" \"npm run watch-views\" \"npm run watch-ts\" \"npm run watch-node\"",
     "start:feature": "set -a; source feature.env; set +a && npm run build && npm start",
     "start-mocks": "node mock-server/start.js",
-    "test": "jest --runInBand --testPathPattern=/test/.*",
+    "test:ci": "jest --runInBand --testPathPattern=/test/.*",
+    "test": "jest --testPathPattern=/test/.*",
```
andrewrlee, line 22: the isolated modules change massively reduces memory usage as well as speeding up tests.
This reduced memory requirements mean that locally we can run the tests in parallel - this combined means the tests run much faster: it knocks over a minute off the test run on my laptop.

package.json
```diff
@@ -95,85 +104,84 @@
...
     "nodemon": "^3.0.1",
     "pdf-parse": "^1.1.1",
-    "prettier": "^2.7.1",
-    "supertest": "^6.3.1",
-    "ts-jest": "^29.0.3",
-    "typescript": "^4.9.3"
+    "prettier": "^3.0.0",
+    "supertest": "^6.3.3",
+    "ts-jest": "^29.1.1",
+    "typescript": "^5.1.6"
```
andrewrlee, line 184: major upgrades hadn't been done for a long time

server/app.js
```diff
@@ -6,7 +6,7 @@ const express = require('express')
 const path = require('path')
 const flash = require('connect-flash')
 const session = require('express-session')
-const ConnectRedis = require('connect-redis')
+const RedisStore = require('connect-redis').default
```
andrewrlee, line 9: as this is js and uses `require`, you have to explicitly reference the default export

tsconfig.json
```diff
@@ -6,7 +6,6 @@
     "moduleResolution": "node",
     "outDir": "./dist",
     "sourceMap": true,
-    "suppressImplicitAnyIndexErrors": true,
```
andrewrlee, line 9: this is a deprecated option so tsc was complaining but we weren't using it anyway

licences-specs/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
 
 task mockTest(type: Test) {
   systemProperty 'geb.env', 'chromeHeadless'
+  systemProperty 'webdriver.chrome.driver', '/usr/local/bin/chromedriver/chromedriver'
```
andrewrlee, line 39: this could cause an issue when running locally but seems like webdriver isn't added to the path anymore in ci.

helm_deploy/values-dev.yaml
```diff
@@ -47,3 +47,6 @@ env:
   LICENCE_RISK_MANAGEMENT_VERSION: '3'
   CURFEW_ADDRESS_REVIEW_VERSION: '2'
   LICENCE_POSTPONE_VERSION: '2'
+
+generic-prometheus-alerts:
+  alertSeverity: cvl-alerts-non-prod
```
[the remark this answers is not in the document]
andrewrlee: Thanks but it's not the same thing as the slack channel, it's the severity level linked to the alert receiver

as seen [here](https://github.com/ministryofjustice/cloud-platform/issues/4164#issuecomment-1282263796)

justin-gardiner (approved):
lgtm

PR #1493, opened 2023-08-10
CVSL-1271 update reject and reinstate proposed address with manageInTheCommunity and associated reasons

louisenorris:


andrewrlee (approved):
LGTM

PR #1494, opened 2023-08-10
CVSL-1271 update reject and reinstate proposed address with hasConsideredChecks

louisenorris:
If a COM flags that an address is unsuitable in the risk section for any reason effectively rejecting the address then sends the case back to the CA. The CA then proposed a new address and returns the case to the COM. The risk section questions which relate the the previous address should then be cleared out as the COM should answer these again considering the new proposed address. The hasConsideredChecks Mandatory address checks question was previously not cleared but now is with this fix. This therefore also means the Mandatory address checks warning is displayed in the task list:

Screenshot 2023-08-10 at 5 00 54 pm

### DIFF
--- a/server/services/licenceService.ts
+++ b/server/services/licenceService.ts
@@ -460,6 +460,7 @@ export class LicenceService {
             'unsuitableReason',
             'manageInTheCommunity',
             'manageInTheCommunityNotPossibleReason',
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

PR #1501, opened 2023-09-07
CVSL-1291 add offender number as a case identifier to emails

louisenorris:
Added `offenderNo` as a case identifier to pass to notify emails in HDC. Templates will need to be updated in Notify but have tested changes work in a new test template in the service. Also put together a basic flow diagram of how the notification functionality is structured in the service to help me do this and included this below in case useful for the PR:

HDC email notifications

From what I can see the `sendReminder` and `sendReminders` functions in the `reminderService` aren't actually used anywhere(?) so might remove them in the future but didn't want to risk doing this now as part of this work.

### DIFF
--- a/server/routes/send.js
+++ b/server/routes/send.js
@@ -29,13 +29,11 @@ module.exports =
       asyncMiddleware(async (req, res) => {
         const { destination, bookingId } = req.params
         const { token, licence, prisoner } = res.locals
-        const { offenderNo } = prisoner
         const transition = transitionsForDestinations[destination]
 
         await notificationService.send({
           transition,
           bookingId,
-          offenderNo,
           token,
           licence,
           prisoner,
--- a/server/routes/send.js
+++ b/server/routes/send.js
@@ -29,11 +29,13 @@ module.exports =
       asyncMiddleware(async (req, res) => {
         const { destination, bookingId } = req.params
         const { token, licence, prisoner } = res.locals
+        const { offenderNo } = prisoner
         const transition = transitionsForDestinations[destination]
 
         await notificationService.send({
           transition,
           bookingId,
+          offenderNo,
           token,
           licence,
           prisoner,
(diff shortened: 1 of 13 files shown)


server/routes/send.js
```diff
@@ -29,11 +29,13 @@ module.exports =
       asyncMiddleware(async (req, res) => {
         const { destination, bookingId } = req.params
         const { token, licence, prisoner } = res.locals
+        const { offenderNo } = prisoner
         const transition = transitionsForDestinations[destination]
 
         await notificationService.send({
           transition,
           bookingId,
+          offenderNo,
```
andrewrlee, line 38: is the offenderNo not already passed through as part of the prisoner below?  
Maybe you don't need to pass it separately?

server/services/notifications/roNotificationHandler.js
```diff
@@ -85,7 +85,7 @@ module.exports = function createRoNotificationHandler(
   }
 
   return {
-    async sendRo({ transition, bookingId, token, user }) {
+    async sendRo({ transition, bookingId, offenderNo, token, user }) {
```
andrewrlee, line 88: prisoner is passed through to here, its just not being deconstructed

server/routes/send.js
louisenorris: Yep - reverted this in latest commit

types/licences.d.ts
```diff
@@ -43,9 +43,20 @@ export interface RoContactDetailsService {
...
 
+interface PrisonerDetails {
+  firstName: string
+  lastname: string
+  offenderNo: stiring
+}
+
+interface Prisoner extends PrisonerDetails {
+  any
+}
```
louisenorris, line 54: Not sure if this is typescript specific enough but the whole prisoner object looks quite extensive and contains another nested objects so wondering if this is ok to just make sure we have type safety on the prisoner parameters we require for the emails?

andrewrlee: this doesn't look like valid typescript to me?

Yes, its fine just to add the fields you are interested in.. not sure why you need the separate interface though?  Could you just have one?

louisenorris: Thanks, corrected this in latest commit

andrewrlee (approved):
LGTM

PR #1531, opened 2024-02-29
CVSL-1729 add migrations

louisenorris:


louisenorris:
I haven't updated the `saveApprovedLicenceVersion` function with the new `v_licence_versions_excluding_deleted` view as think this is not necessary. The new view will show the saved versions as long as the `deleted_at` timestamp is `NULL` so think it's best to save in the original table but wanted to highlight and check this.

andrewrlee (approved):
LGTM

PR #1532, opened 2024-03-04
CVSL-1706 move reset code to soft delete instead

louisenorris:


server/data/licenceClient.ts
```diff
@@ -206,6 +214,16 @@ export class LicenceClient {
...
     await db.query(query)
   }
+
+  async softDeleteLicence(bookingId: number): Promise<void> {
+    const query = {
+      text: 'UPDATE v_licences_excluding_deleted SET deleted_at = current_timestamp where booking_id = $1 and deleted_at is null;',
+      values: [bookingId],
+    }
+
+    await db.query(query)
```
andrewrlee, line 224: can you wrap these two calls in a transaction? so they succeed or fail together? 

if its not easy to then we can have a chat about it

andrewrlee (approved):
LGTM

PR #1533, opened 2024-03-27
CVSL-1778 update dependencies

andrewrlee:


.husky/_/husky.sh
```diff
@@ -1,36 +0,0 @@
...
-  if [ $exitCode != 0 ]; then
-    echo "husky - $hook_name hook exited with code $exitCode (error)"
-  fi
-
-  if [ $exitCode = 127 ]; then
-    echo "husky - command not found in PATH=$PATH"
-  fi
-
-  exit $exitCode
-fi
```
louisenorris, line 36: ooo assume we do need this with a newer version of husky?

louisenorris (approved):
LGTM

.husky/_/husky.sh
andrewrlee: husky deleted it itself!

PR #1540, opened 2024-05-02
CVSL-1814 fix admin search

louisenorris:
Fix admin search to also find soft deleted licences by searching the licences table instead of the view which excludes soft deleted licences.


server/services/licenceSearchService.js
```diff
@@ -15,14 +15,14 @@ module.exports = function createLicenceSearchService(
   nomisClientBuilder,
   prisonerSearchApi
 ) {
-  const bookingIdForExistingLicence = async (bookingId) => {
-    const licence = await licenceClient.getLicence(bookingId)
+  const bookingIdForLicenceIncludingSoftDeleted = async (bookingId) => {
```
louisenorris, line 18: renamed to make it clear that search results could include soft deleted licences

louisenorris, line 18: renamed to make it clear that search results could include soft deleted licences

server/services/licenceSearchService.js
```diff
@@ -15,14 +15,14 @@ module.exports = function createLicenceSearchService(
...
+    const licence = await licenceClient.getLicenceIncludingSoftDeleted(bookingId)
```
louisenorris, line 19: this licence search was only being used by the admin search so switched the `getLicence` client call to a new `getLicenceIncludingSoftDeleted` client call which searched the `licences` table instead of the view which excludes soft deleted licences.

SebNegahbanMoJ (approved):

PR #1548, opened 2024-06-11
CVSL-1935 add new ineligible reasons

louisenorris:


test/services/nomisPushService.test.js
```diff
@@ -85,6 +85,31 @@ describe('nomisPushService', () => {
           data: { type: 'excluded', status: 'Yes', reason: 'deportation' },
           approvalStatus: { approvalStatus: 'INELIGIBLE', refusedReason: 'FNP' },
         },
+        {
+          example: 'Excluded - cja2003_19ZA',
+          data: { type: 'excluded', status: 'Yes', reason: 'cja2003_19ZA' },
+          approvalStatus: { approvalStatus: 'INELIGIBLE', refusedReason: 'CJA2003_19ZA' },
+        },
```
SebNegahbanMoJ, line 92: This one doesn't seem to be listed anywhere else, has it been added previously and this is just adding it to a spec?

louisenorris: Ah!, Forgot to add a comment for this sorry, it was missing from the test so added it in, it's one of the current reasons and just seemed weird that it was the only one left out of the test for ineligible reasons 😅

SebNegahbanMoJ (approved):

PR #1560, opened 2024-07-11
CVSL-2019 COM caselist fix

louisenorris:


server/services/roService.ts
```diff
@@ -22,8 +24,25 @@ export class RoService {
...
+        const sortedSentences = sentences.sort((a, b) => {
+          return moment(
+            a.sentenceDetail.topupSupervisionExpiryCalculatedDate || a.sentenceDetail.licenceExpiryCalculatedDate
+          ).diff(b.sentenceDetail.topupSupervisionExpiryCalculatedDate || b.sentenceDetail.licenceExpiryCalculatedDate)
+        })
+        const latestSentence = sortedSentences.pop()
+        latestOffenderSentences.push(latestSentence)
+      }
+    })
+    return latestOffenderSentences
```
louisenorris, line 45: New criteria for showing the latest sentences is:
- show the sentence with the latest calculated TUSED if present
- if not show the sentence with the latest calculated LED
- if neither of these dates are present then we don't think we are able to identify the latest sentence so show all the sentences for that offender (this should be a rare edge case and showing all sentences is current behaviour anyway).

server/services/roService.ts
```diff
@@ -22,8 +24,25 @@ export class RoService {
...
-  private filterMostRecentActiveBooking(offenderSentences: OffenderSentence[]): OffenderSentence[] {
-    return offenderSentences.filter((o) => o.mostRecentActiveBooking === true)
+  getLatestSentences(offenderSentences: OffenderSentence[]): OffenderSentence[] {
+    const groupedOffenderSentencess = groupBy(offenderSentences, ({ offenderNo }) => offenderNo)
+    let latestOffenderSentences = []
+
+    Array.from(groupedOffenderSentencess.values()).flatMap((sentences) => {
+      const hasNoDatesToCompare = sentences.find((b) => absentDatesToCompare(b))
+      if (hasNoDatesToCompare) {
+        sentences.forEach((sentence) => latestOffenderSentences.push(sentence))
```
andrewrlee, line 34: **todo:** `flatmap` is used to flatten a list of responses which means we can avoid mutation.
The way it's used here currently is to solely do the iteration (it could be replaced by a `forEach`)  

to use `flatMap` more effectively you'd assign the result of the flatMap:

`const latestOffenderSentences  = Array.from(groupedOffenderSentencess.values()).flatMap((sentences)...`

and then in each branch would return an array you want to flatten into the single returned array
so on this line:  `return sentences`
and in the other branch: `return [sortedSentences.pop()]`

server/services/roService.ts
```diff
@@ -102,3 +121,9 @@ function toResponsibleOfficer(offenderNumber: string, offenderManager: Community
     probationAreaDescription: probationArea.description,
   }
 }
+
+function absentDatesToCompare(booking): Boolean {
```
andrewrlee, line 125: **suggestion(non-blocking):** like the name above `hasNoDatesToCompare` could use that here: `hasNoDatesToCompare`

louisenorris: Changing it to `noDatesToCompare` to avoid `'hasNoDatesToCompare' is already declared in the upper scope` error

server/services/roService.ts
louisenorris: Thank you. Refactored to use `flatMap` more effectively as suggested. Stupid question please - why do we return the last sorted sentence (`sortedSentences.pop()`) in an array?

andrewrlee (approved):
LGTM

PR #1562, opened 2024-07-12
Revert "Revert "CVSL-1967 revert ro case list most recent booking filter""

andrewrlee:
Reverts ministryofjustice/licences#1556


louisenorris (approved):
LGTM

PR #1571, opened 2024-11-28
CVSL-2291 wip

louisenorris:


server/routes/taskList.ts
```diff
@@ -128,23 +127,9 @@ export = (
...
-
-          if (errorCode) {
-            return res.render('taskList/taskListBuilder', {
-              ...model,
-              taskListModel: getTasksForBlocked(errorCode),
-              errors: [],
-            })
-          }
-        }
+        const errorCode = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
```
andrewrlee, line 130: So this would make an extra api call for every tasklist where in quite a lot of cases it'd be irrelevant.. in other languages we could make this lazy and pass this through - only calling it when necessary but in JS using async await it means we'd have to change the [colour of the function](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/#:~:text=Red%20functions%20are%20asynchronous%20ones,just%20made%20a%20red%20function.) we're calling to also it to do async await also. 

One way around this could be to split out a getCaTasklist in taskListModels,
so here we:
```js
if (role === 'CA') {
  const errorCode = await caService.getReasonForNotContinuing(bookingId, res.locals.token)
  return res.render('taskList/taskListBuilder', {
              ...model,
              taskListModel: getCaTaskList(licenceStatus, errorCode, licence || {}),
              errors: [],
            })
}

const taskListModel = getTaskListModel(req.user.role, postRelease, licenceStatus, licence || {})

        return res.render('taskList/taskListBuilder', {
          ...model,
          taskListModel,
          errors: [],
        })
```

server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -22,30 +22,41 @@ const eligibilityTask = namedTask('eligibilityTask')
...
     const { optedOut, eligible, bassReferralNeeded, addressUnsuitable } = decisions
-    const { eligibility, optOut } = tasks
+    const { eligibility, optOut, bassRequest } = tasks
 
     const eligibilityDone = eligibility === 'DONE'
     const optOutUnstarted = optOut === 'UNSTARTED'
     const optOutRefused = optOut === 'DONE' && !optedOut
 
+    // completed only when both curfewAddress & bassRequest are DONE (once one is done all sections curfew/CAS/opt out are done) and not opted out
+    const isCurfewAddressCompleted = optOutRefused && tasks.curfewAddress === 'DONE' && bassRequest === 'DONE'
```
andrewrlee, line 34: mmh.. does this work? - isn't BASS request only DONE if it's a BASS? 

is the logic to determine whether the case should be handed over already present - isn't it when either?:
* eligibleForBASSReview: `optOutRefused && bassReferralNeeded && allowedTransition !== 'caToDmRefusal']`
* eligibleForAddressReview: `optOutRefused && !bassReferralNeeded && allowedTransition !== 'caToDmRefusal'`

so:
```ts
const eligibleForHandover = eligibleForBASSReview || eligibleForAddressReview

[submitBassReview, eligibleForBassReview && !errorCode],
[submitAddressReview, eligibleForAddressReview && !errorCode],
[caBlocked, eligibleForHandover && errorCode]
```

server/services/licence/licenceStatus.ts
```diff
@@ -82,6 +82,7 @@ const getInitialState = (stage, licenceRecord) => {
       offenderIsMainOccupier,
       curfewAddressRejected,
       addressUnsuitable,
+      curfewAddressProposed,
```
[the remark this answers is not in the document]
louisenorris: Totally 😄

server/routes/viewModels/taskLists/caTasks.js
louisenorris: So bassRequest === 'DONE' is populated at the point a user selects either

louisenorris: So `bassRequest === 'DONE'` is populated at the point a user selects either provided an address, CAS2 or opted out. As soon as one is selected they are all marked `DONE`.

Screenshot 2024-12-13 at 12 57 22

louisenorris: But we'll look into your suggestion above more 👀

louisenorris: Think we need to do a bit of testing on it 🧪

louisenorris: Updated block logic following above suggestion 😄

andrewrlee (approved):
LGTM assuming its just one opt out

PR #1573, opened 2024-12-05
CVSL-2330 add licenceInCvl to client call, service and status decision

louisenorris:


server/views/admin/licences/activity.pug
```diff
@@ -8,6 +8,10 @@ mixin detail(type, details)
         + notify(details)
     when 'RESET'
         | Licence has been reset
+    when 'COMPLETE_IN_HDC'
+        | Licence to be completed in HDC
+    when 'COMPLETE_IN_CVL'
+        | Licence to be completed in CVL
```
louisenorris, line 14: Looks like this in the service: 
Screenshot 2024-12-09 at 15 35 44

server/data/licenceClient.ts
```diff
@@ -230,6 +239,16 @@ export class LicenceClient {
     await db.query(query)
   }
 
+  async setLicenceInCvl(licence_in_cvl: boolean, bookingId: number): Promise<void> {
+    await db.inTransaction(async (client) => {
+      await client.query({
+        text: 'UPDATE v_licences_excluding_deleted SET licence_in_cvl = $1 where booking_id = $2',
+        values: [licence_in_cvl, bookingId],
+      })
+      return setLicenceVersionInCvl(licence_in_cvl, bookingId)
```
andrewrlee, line 248: mmh.. think the client is the transactional bit - so needs to be passed into the setLicenceVersionInCvl to continue the transaction?

(wonder this is wrong elsewhere)

andrewrlee: probably need to look at this deeper...

server/views/admin/licences/activity.pug
andrewrlee: meh.. could we call have the details be "licence to be created in.." 

(don't worry about the routes and page names etc...)

server/data/licenceClient.ts
andrewrlee:
```
export const inTransaction = async (callback) => {
  const client = await pool.connect()
  try {
    await client.query('BEGIN')
    const result = await callback(client)
    await client.query('COMMIT')
    return result
  } catch (error) {
    await client.query('ROLLBACK')
    throw error
  } finally {
    client.release()
  }
}
```

yes - I think they all need to be done on the same client instance

andrewrlee (approved):
LGTM

PR #1578, opened 2025-01-09
CVSL-2292 add licence client call

louisenorris:


server/data/licenceClient.ts
```diff
@@ -194,6 +194,20 @@ export class LicenceClient {
...
+      text: `select l.booking_id from v_licences_excluding_deleted l
+                    where stage = $1
+                    and licence -> 'proposedAddress' -> 'optOut' ->> 'decision' = 'No'
+                    and ((licence -> 'bassReferral' -> 'bassRequest' ->> 'bassRequested' = 'Yes' and licence -> 'proposedAddress' -> 'addressProposed' ->> 'decision' = 'No')
+                    or (licence -> 'bassReferral' -> 'bassRequest' ->> 'bassRequested' = 'No' and licence -> 'proposedAddress' -> 'addressProposed' ->> 'decision' = 'Yes'))`,
+      values: [stage],
+    }
+
+    const { rows } = await db.query(query)
+    return rows
```
andrewrlee, line 208: meh.. think we usually have simple tests for this but it's a bit of double account so NBD

server/data/probationSearchApi.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const { apiUrl } = config.probationSearchApi
+
+const agentOptions = {
+  maxSockets: config.probationSearchApi.agent.maxSockets,
+  maxFreeSockets: config.probationSearchApi.agent.maxFreeSockets,
+  freeSocketTimeout: config.probationSearchApi.agent.freeSocketTimeout,
+}
+
+const batchRequests = async (args, batchSize, call) => {
```
andrewrlee, line 19: **suggestion**: move to util code and share across prison and probation? 

This seems to be missing type info

server/routes/licencesRequiringComAssignmentReport.ts
```diff
@@ -0,0 +1,25 @@
+import { asyncMiddleware } from '../utils/middleware'
+
+export = (licenceSearchService) => (router) => {
+  router.get(
+    '/',
+    asyncMiddleware(async (req, res) => {
+      return res.render('licencesRequiringComAssignmentReport', {})
```
andrewrlee, line 7: **suggestion**: add audit to these actions? 
this template should probably not be in the root of `views`

Do you need this though - we don't really know how this is going to be linked but maybe we just need to render the report on a GET for now?

server/routes/licencesRequiringComAssignmentReport.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  router.post(
+    '/',
+    asyncMiddleware(async (req, res) => {
+      const licencesRequiringComAssignment = await licenceSearchService.getLicencesRequiringComAssignment(
+        req.user.username,
+        res.locals.user.activeCaseLoadId
+      )
+      res.contentType('text/csv')
+      res.set('Content-Disposition', `attachment;filename=HDC12 no COM assigned.csv`)
```
andrewrlee, line 19: **suggestion:** talk to content design for a file name, might want to include the prison id and date/time in there

server/services/licenceSearchService.js
```diff
@@ -88,5 +122,37 @@ module.exports = function createLicenceSearchService(
...
       return getlicencesCSV(prisonerDecoratedLicences)
     },
+
+    async getLicencesRequiringComAssignment(username, prisonId) {
+      const systemToken = await signInService.getClientCredentialsTokens(username)
+      const licencesInStageWithAddressOrCasLocation = await licenceClient.getLicencesInStageWithAddressOrCasLocation(
+        'ELIGIBILITY',
+        systemToken
+      )
+      const bookingIds = await licencesInStageWithAddressOrCasLocation.map((l) => l.booking_id)
```
andrewrlee, line 132: don't need to await synchronous operations

server/services/licenceSearchService.js
```diff
@@ -88,5 +122,37 @@ module.exports = function createLicenceSearchService(
...
+      const prisoners = await prisonerSearchApi(systemToken).getPrisoners(bookingIds)
+      const prisonersFilteredByPrisonCloseToHdced = await prisoners.filter(
+        (p) =>
+          p.prisonId === prisonId &&
+          moment(p.homeDetentionCurfewEligibilityDate, 'YYYY-MM-DD').isBetween(moment(), moment().add(14, 'weeks'))
+      )
+      const offenderNumbers = await prisonersFilteredByPrisonCloseToHdced.map((p) => p.prisonerNumber)
+      const probationDetails = await probationSearchApi(systemToken).getPersonProbationDetails(offenderNumbers)
+      const probationDetailsWithUnallocatedCom = await probationDetails.filter((pd) => {
+        const results = pd.offenderManagers.some(
```
andrewrlee, line 142: **suggestion:** extract this into a separate function so you can is
`probationDetails.filter((pd) => isUnallocated(pd))`

test/services/licenceSearchService.test.js
```diff
@@ -33,34 +39,106 @@ describe('licenceSearchService', () => {
...
         },
         {
           bookingId: '3',
           prisonerNumber: 'AAAA13',
+          firstName: 'Tim',
+          lastName: 'North',
           prisonId: 'MDI',
           prisonName: 'Moorland (HMP & YOI)',
-          homeDetentionCurfewEligibilityDate: '01-01-2021',
+          homeDetentionCurfewEligibilityDate: moment().add(13, 'weeks').format('YYYY-MM-DD'),
```
andrewrlee: Think we need to be test closer to fence posts here.. 
Would be good to have cases to cover day before, exactly 14 weeks, day after.

server/services/licenceSearchService.js
```diff
@@ -70,6 +71,39 @@ module.exports = function createLicenceSearchService(
...
+  const getPrisonerProbationDecoratedLicences =
+    ({ prisoners, probationDetails }) =>
+    (licencesAcc, licence) => {
+      const prisoner = prisoners.find((p) => p.bookingId === licence.booking_id.toString())
+      if (!prisoner || prisoner.status === 'INACTIVE OUT') return licencesAcc
+
+      const probationDetail = probationDetails.find((pd) => pd.otherIds.nomsNumber === prisoner.prisonerNumber)
+      if (!probationDetail) return licencesAcc
+      return [
+        ...licencesAcc,
```
andrewrlee, line 83: **suggeston:** this gets very slow for large numbers of records and the reducer is a bit complicated.. 

Is it ever possible to not have a probation details here?  
* if it is not possible you could get away with `map` and maybe just assert that you find a match 
* if it is possible you could do a flatMap where you return `[]` if there no match or `[{ .... record ....}]` if there is.

server/routes/licencesRequiringComAssignmentReport.ts
louisenorris: Yeah, just needed to render it to get the report working for now. Currently in the root of the `views` as I didn't know where to put it just yet. Is it ok to leave it there temporarily?

server/routes/licencesRequiringComAssignmentReport.ts
louisenorris: That's the file name stated on the ticket 😅 will check with Design re date and time 👍

server/data/probationSearchApi.ts
louisenorris: Moved to utils as suggested and added types to parameters

server/routes/licencesRequiringComAssignmentReport.ts
louisenorris: Audit for action added.

server/routes/licencesRequiringComAssignmentReport.ts
louisenorris: Thanks for confirming the new report title, just to double check format - so an example of `yyyy-MM-dd_HHmm` would be `2025-01-Th_1120` (Th being Thursday)

louisenorris: Updated to date format to `YYYY-MM-DD_HHmm` eg `2025-01-16_1120`

server/services/licenceSearchService.js
```diff
@@ -82,11 +125,33 @@ module.exports = function createLicenceSearchService(
...
+    async getLicencesRequiringComAssignment(username, prisonId) {
+      const systemToken = await signInService.getClientCredentialsTokens(username)
+      const licencesInStageWithAddressOrCasLocation = await licenceClient.getLicencesInStageWithAddressOrCasLocation(
+        'ELIGIBILITY',
+        systemToken
+      )
+      const bookingIds = licencesInStageWithAddressOrCasLocation.map((l) => l.booking_id)
+      const prisoners = await prisonerSearchApi(systemToken).getPrisoners(bookingIds)
+      const prisonersFilteredByPrisonCloseToHdced = prisoners.filter(
+        (p) => p.prisonId === prisonId && isCloseToHdced(p)
```
andrewrlee, line 144: could you exclude `prisoner.status === 'INACTIVE OUT'` prisoners here too?

louisenorris: Moved as suggested, thanks!

andrewrlee (approved):
LGTM

PR #1579, opened 2025-01-17
CVSL-2388 add service layer functionality

louisenorris:


server/services/licenceSearchService.js
```diff
@@ -128,30 +154,38 @@ module.exports = function createLicenceSearchService(
       const bookingIds = licencesWithCOM.map((l) => l.booking_id)
       const prisoners = await prisonerSearchApi(systemToken).getPrisoners(bookingIds)
       const licencesAcc = []
-      const prisonerDecoratedLicences = licencesWithCOM.reduce(getPrisonerDecoratedLicences(prisoners), licencesAcc)
-      return getlicencesCSV(prisonerDecoratedLicences)
+      const decoratedLicences = licencesWithCOM.reduce(getPrisonerDecoratedLicences(prisoners), licencesAcc)
+      return getPrisonerDecoratedLicencesCSV(decoratedLicences)
```
louisenorris, line 158: **note:** tried to improve some naming also

server/services/licenceSearchService.js
```diff
@@ -43,6 +43,43 @@ module.exports = function createLicenceSearchService(
     }
   }
 
+  const isCloseToHdced = (p) => {
+    return moment(p.homeDetentionCurfewEligibilityDate, 'YYYY-MM-DD').isBetween(
+      moment(),
+      moment().add(14, 'weeks'),
+      undefined,
+      '[]'
```
louisenorris, line 51: **note:** last parameter of `'[]'` required to make `isBetween` comparison dates of now and 14 weeks inclusive. Tests before this were flaky.

server/views/comAssignedLicencesForHandoverReport.pug
```diff
@@ -0,0 +1,20 @@
+extends layout
```
andrewrlee, line 1: you could create a `reports` folder inside `views/` and move all three reports in there. 

The `/` of view is for the very generic stuff which is shared across everything

test/services/licenceSearchService.test.js
```diff
@@ -10,7 +10,8 @@ let prisonerSearchAPI
 let probationSearchApi
 let restPrisonerClientBuilder
 let restProbationClientBuilder
-const hdced = moment().add(14, 'weeks').subtract(1, 'days').format('DD-MM-YYYY')
+const hdcedWithin14Weeks = moment().add(14, 'weeks').subtract(1, 'days').format('DD-MM-YYYY')
+const hdced14Weeks = moment().add(14, 'weeks').format('DD-MM-YYYY')
```
andrewrlee, line 14: test for hdced + 14 weeks + 1 day? 

Good to test all around the boundary

server/services/licenceSearchService.js
```diff
@@ -135,30 +156,38 @@ module.exports = function createLicenceSearchService(
...
-        systemToken
-      )
-      const bookingIds = licencesInStageWithAddressOrCasLocation.map((l) => l.booking_id)
-      const prisoners = await prisonerSearchApi(systemToken).getPrisoners(bookingIds)
-      const prisonersFilteredByPrisonCloseToHdced = prisoners.filter(
-        (p) => p.prisonId === prisonId && p.status !== 'INACTIVE OUT' && isCloseToHdced(p)
-      )
-      const offenderNumbers = prisonersFilteredByPrisonCloseToHdced.map((p) => p.prisonerNumber)
-      const probationDetails = await probationSearchApi(systemToken).getPersonProbationDetails(offenderNumbers)
+      const licences = await getEligibleLicencesWithAddressOrCasLocation(username)
```
andrewrlee, line 164: This file has grown a lot - It'd probably make sense to add a reports service.. this functionality isn't really anything to do with search

server/app.js
```diff
@@ -441,6 +442,10 @@ module.exports = function createApp({
     '/hdc/licencesRequiringComAssignment',
     secureRoute(licencesRequiringComAssignmentRouter(licenceSearchService, audit))
   )
+  app.use(
+    '/hdc/comAssignedLicencesForHandover',
+    secureRoute(comAssignedLicencesForHandoverRouter(licenceSearchService, audit))
```
andrewrlee, line 447: this should sit with admin routes above, the other admin report has a path: `'/admin/downloadCasesWithCOM/',`

andrewrlee: should probably sit in `/admin` routes folder too

andrewrlee: This will also restrict the route so only BATCHLOAD can access it. 
`/server/routes/config/authorisation.js`

andrewrlee: We're probably missing an entry in `authorisation.js` to limit the CA report to only CAs.

louisenorris: Oh but this report is the other CA one, the new admin report is to follow in the next ticket?

server/views/comAssignedLicencesForHandoverReport.pug
louisenorris: Maybe admin reports should be in a sub `reports` folder of the `admin` folder 🤔

andrewrlee: yes sounds good

server/app.js
louisenorris: Limited both CA reports to only CAs in `authorisation.js`

server/data/licenceClientTypes.ts
```diff
@@ -21,14 +21,14 @@ export interface Case {
   booking_id: number
   stage: LicenceStage
   version: number
+  transition_date: Date
```
louisenorris, line 24: added reportsService in typescript and needed `transition_date` as part of `Case` type so moved to `Case` to represent the db schema

server/data/probationSearchApiTypes.ts
```diff
@@ -29,7 +29,7 @@ declare module 'probationSearchApi' {
       probationArea: Schemas['ProbationArea']
 
       /** Offender Manager Active */
-      active?: string
+      active?: boolean
```
louisenorris, line 32: corrected

test/services/licenceService.test.ts
```diff
@@ -31,6 +32,7 @@ describe('licenceService', () => {
         booking_id: 1,
         stage: undefined,
         version: 2,
+        transition_date: moment('2018-05-31 15:23:39').toDate(),
```
louisenorris, line 35: needed to add `transition_date` here and below as moved it into the `Case` type

gurps-singh (approved):

types/licences.d.ts
```diff
@@ -140,8 +140,12 @@ export interface LduService {
 
 export interface LicenceSearchService {
   findForId: (username: string, id: string) => Promise<number | undefined>
+}
+
+export interface ReportsService {
```
andrewrlee, line 145: If you've defined ReportService as typescript you shouldn't need to define an interface for it - these are just to help typescript work with js files

andrewrlee: ah, think maybe it's because you've defined the service as a function that returns an object..

If you defined it as a class you'd get this for free. 

To avoid having to redeclare all the functions you could also us the `ReturnType` 
like how we do for dataAccess in any typescript repo:
```ts
const dataAccess = () => ...
export type DataAccess = ReturnType<typeof dataAccess>
```

andrewrlee (approved):
LGTM

PR #1580, opened 2025-01-20
CVSL-2292 enable com unallocated block in preprod, map probation sear…

louisenorris:
…ch api, return empty string in report when no pdu


server/services/licenceSearchService.js
```diff
@@ -79,7 +79,12 @@ module.exports = function createLicenceSearchService(
   }
 
   const isCloseToHdced = (p) => {
-    return moment(p.homeDetentionCurfewEligibilityDate, 'YYYY-MM-DD').isBetween(moment(), moment().add(14, 'weeks'))
+    return moment(p.homeDetentionCurfewEligibilityDate, 'YYYY-MM-DD').isBetween(
+      moment(),
+      moment().add(14, 'weeks'),
+      undefined,
+      '[]'
```
louisenorris, line 86: **note:** realised I needed the final parameter of '[]' which is required to make isBetween comparison dates of now and 14 weeks inclusive.

test/services/licenceSearchService.test.js
```diff
@@ -153,6 +163,21 @@ describe('licenceSearchService', () => {
...
           ],
         },
+        {
+          otherIds: {
+            nomsNumber: 'AAAA17',
+          },
+          offenderManagers: [
+            {
+              active: true,
+              probationArea: {},
```
louisenorris, line 173: Added to test handling no PDU in report

server/services/licenceSearchService.js
```diff
@@ -94,7 +99,7 @@ module.exports = function createLicenceSearchService(
           prisonerFirstname: prisoner.firstName,
           prisonLastname: prisoner.lastName,
           HDCED: moment(prisoner.homeDetentionCurfewEligibilityDate).format('DD-MM-YYYY'),
-          PDU: probationDetail.offenderManagers[0].probationArea.description,
+          PDU: probationDetail ? probationDetail.offenderManagers[0].probationArea.description : '',
```
andrewrlee, line 102: Should we look for an active offender manager? 
might be better to do `?.[0]` and then `?.` to handle no offender managers rather than blow up?

louisenorris: optional chaining causing issues with linting in a javascript file so opted for using destructuring to get description value.

gurps-singh (approved):

server/services/licenceSearchService.js
louisenorris: Also added a look up for the active community offender manager in this function 😅

PR #1581, opened 2025-01-22
CVSL-2301 hide continue button and add label text when licence is to made in CVL

gurps-singh:
This PR is to customise the create licence task list based on whether the licence is to be created in CVL. 

Licence to be created in HDC
Screenshot 2025-01-22 at 10 07 22

Licence to be created in CVL
Screenshot 2025-01-22 at 10 04 37

### DIFF
--- a/server/routes/viewModels/taskLists/caTasks.js
+++ b/server/routes/viewModels/taskLists/caTasks.js
@@ -140,7 +140,6 @@ module.exports = {
         dmRefused,
         eligible,
         postponed,
-        useCvlForLicenceCreation,
       } = decisions
 
       const { bassOffer } = tasks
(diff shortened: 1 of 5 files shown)


server/routes/viewModels/taskLists/caTasks.js
```diff
@@ -182,8 +181,7 @@ module.exports = {
           validAddress &&
             !['caToDm', 'caToDmRefusal', 'caToRo'].includes(allowedTransition) &&
             !dmRefused &&
-            !postponed &&
-            !useCvlForLicenceCreation,
+            !postponed,
```
andrewrlee, line 184: so.. now show task for these CVL cases?

andrewrlee (approved):
LGTM

server/routes/viewModels/taskLists/caTasks.js
gurps-singh: If they are CVL, the task will now be shown with a label and no continue button (provided they meet the other checks) and if HDC, the task will show without a label and the continue button.

louisenorris (approved):
LGTM

PR #1583, opened 2025-01-22
CVSL-2389 fix cases sitting with COM report auditing

louisenorris:


server/services/reportsService.ts
```diff
@@ -22,38 +30,53 @@ export = function createReportsService(
...
   }
 
-  const getEligibleLicencesWithAddressOrCasLocation = async (): Promise<Case[]> => {
+  private static getPduDescription(probationDetail: OffenderDetail): string {
+    if (!probationDetail) return ''
+    const activeOffenderManager = probationDetail.offenderManagers.find((om) => om.active)
+    if (!activeOffenderManager) return ''
+    const { description } = activeOffenderManager.probationArea
+    return description || ''
+  }
```
SebNegahbanMoJ, line 46: I think this can probably be reduced/made a bit more concise by turning it into something like the following

```
private static getPduDescription(probationDetail: OffenderDetail): string {
    const activeOffenderManager = probationDetail?.offenderManagers.find((om) => om.active)
    return activeOffenderManager?.probationArea.description || ''
}
  ```
  
There may be a reason that doesn't work, but by carrying the `null` through to the final return check, we don't need to have as many other explicit returns earlier on for similar `null` checks

File: server/services/reportsService.ts [hunk 2]
```diff
@@ -22,38 +30,53 @@ export = function createReportsService(
...
+
+  private static getComName(probationDetail: OffenderDetail): string {
+    if (!probationDetail || ReportsService.isUnallocated(probationDetail)) return ''
+    const activeOffenderManager = probationDetail.offenderManagers.find((om) => om.active)
+    const { forenames, surname } = activeOffenderManager.staff
```
SebNegahbanMoJ, line 51: I think we might be missing a check for if `activeOffenderManager` is null here (if the `find` doesn't return anything? Presumably it always should, but you've accounted for it in `getPduDescription` above, so I assume there's a possibility.

louisenorris: so the check on line 49 will return an empty string if the `probationDetail` exists but is unallocated so if we pass that check I think there will always be an `activeOffenderManager` present? Is that a fair assumption? 😅

server/services/reportsService.ts
louisenorris: refactored, yeah not sure why I did not do this, thank you!

server/services/reportsService.ts
```diff
@@ -7,13 +7,21 @@ import type { OffenderDetail } from 'probationSearchApi'
...
-  probationSearchApi
-) {
-  const isCloseToHdced = (p: Prisoner): boolean => {
+export interface DecorationDetails {
+  licences: Case[]
+  prisoners: Prisoner[]
+  probationDetails: OffenderDetail[]
+}
+
+export class ReportsService {
```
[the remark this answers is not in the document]
louisenorris: made default export as suggested 👍

gurps-singh (approved):

server/services/reportsService.ts
gurps-singh: Think that is a valid assumption to have.

PR #1586, opened 2025-02-10
CVSL-2389 add prison column to admin report csv

louisenorris:


server/services/reportsService.ts
```diff
@@ -16,7 +16,7 @@ export interface DecorationDetails {
 export interface PrisonerProbationRecord {
   prisonerNumber: string
   prisonerFirstname: string
-  prisonLastname: string
+  prisonerLastname: string
```
louisenorris, line 19: **note:** fixed typo

server/services/reportsService.ts
```diff
@@ -149,9 +149,10 @@ export default class ReportsService {
       const pdu = ReportsService.getPduDescription(probationDetail)
       return [
         {
+          prison: prisoner.prisonId,
```
louisenorris, line 152: **note:** only adding prison column to new admin report as the CA reports are specific to a single prison, where as this report is across all prisons so useful to know which prison a prisoner is at.

andrewrlee (approved):
LGTM

PR #1598, opened 2025-05-21
CVSL-2890 sanitise test data

gurps-singh:


licences-specs/src/test/resources/GebConfig.groovy
```diff
@@ -17,9 +17,7 @@ environments {
     driver = {
       ChromeOptions options = new ChromeOptions()
       options.addArguments('headless')
```
gurps-singh, line 19: **note:** ‏matched this to other teams that are using groovy and seems to have fixed the issue of not running tests locally.

AledWEvans (approved):
LGTM - I know the Int test are failing and it's a prior issue but would be good to sort them all the same

PR #1614, opened 2025-07-21
CVSL-3090 remove agency notification forms

louisenorris:


test/routes/forms.test.ts
```diff
@@ -1,57 +0,0 @@
...
-import standardRouter from '../../server/routes/routeWorkers/standardRouter'
-import createFormsRoute from '../../server/routes/forms'
-
-describe('GET /forms', () => {
-  let app
-  beforeEach(() => {
-    app = createApp('caUser')
-  })
-
-  test('gets agency_notification form', () => {
```
louisenorris, line 26: **note:** removed this test file as I believe it was only present to test the `agency_notification` form specifically which has now also been removed. The handler tests require gotenberg end to end integration so test coverage is present via `/licences/integration_test/routes/forms.test.ts`

PR #1617, opened 2025-07-21
CVSL-3090 stop forms from opening in new tabs to comply with accessibility guidance

louisenorris:


AledWEvans (approved):
LGTM - does this work with all browser? I think it should right!

PR #1669, opened 2026-03-23
CVSL-3936 dependency update

louisenorris:


package-lock.json
```diff
@@ -240,7 +240,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
```
louisenorris, line 243: **question:** want to check the removals of the peer tag is ok. They are all removed by the audit fix but haven't yet been able to find what this tag means 👀

AledWEvans (approved):
LGTM

PR #1679, opened 2026-04-20
CVSL-3836 Prog HDC - Migration - Migrate an Active case -- added Knex…

AledWEvans:
added script to create the migration log table!


andrewrlee (approved):
LGTM

PR #1684, opened 2026-05-05
CVSL-2922 env vars

nishanthmahasamudram:


helm_deploy/licences/values.yaml
```diff
@@ -4,6 +4,10 @@ generic-service:
   replicaCount: 4
   productId: "DPS062"
 
+  service:
+    labels:
+      app: licences
```
andrewrlee: don't think this is needed?

helm_deploy/licences/values.yaml
```diff
@@ -4,6 +4,10 @@ generic-service:
...
+
   image:
     repository: quay.io/hmpps/licences
     tag: latest
```
andrewrlee: this needs to be app_version, it gets replaced at deployment time.
We definitely don't want to always deploy the latest

andrewrlee (approved):
LGTM

PR #1704, opened 2026-05-15
CVSL-4095 update to use latest version of gotenberg

andrewrlee:


server/data/gotenbergClient.ts
```diff
@@ -23,7 +23,9 @@ export class GotenbergClient {
     const { headerHtml, footerHtml, marginBottom, marginLeft, marginRight, marginTop } = options
     try {
       const request = superagent
-        .post(`${this.gotenbergHost}/convert/html`)
+        .post(`${this.gotenbergHost}/forms/chromium/convert/html`)
+        .set('Content-Type', 'multi-part/form-data')
+        .buffer(true)
```
andrewrlee, line 28: there's an "javascript integration" test which uses a real version of gotenberg that is deployed to circle ci - so this should test this code

andrewrlee: this test has been detecting issues but is now passing..

nishanthmahasamudram (approved):
LGTM

PR #1713, opened 2026-06-01
CVSL-4136 fix reporting instructions validation

louisenorris:


server/routes/reporting.js
```diff
@@ -1,11 +1,78 @@
...
+    const standard = createStandardRoutes({ formConfig, licenceService, sectionName })
+
+    function get(req, res) {
+      const { bookingId, action } = req.params
+      const errorObject = req.flash('errors')[0] || {}
+      const userInput = req.flash('userInput')[0] || {}
+      return formGet(req, res, 'reportingInstructions', bookingId, errorObject, userInput, action)
+    }
+
+    function formGet(req, res, formName, bookingId, errorObject, userInput, action) {
```
andrewrlee: we can simplify this by merging the pairs of methods:

`formGet` and `get` => `get`
 `formPost` and `post` => `post`

server/routes/reporting.js
```diff
@@ -1,11 +1,78 @@
...
+        postRelease: res.locals.postRelease,
+      })
+
+      await standard.pushStatus(updatedLicence, formName, bookingId, req.user.username)
+
+      if (req.body.anchor) {
+        return res.redirect(`${nextPath}${bookingId}#${req.body.anchor}`)
+      }
+
+      if (req.body.path) {
```
andrewrlee: do these hidden fields (`anchor`, `path`) get submitted via the pug form? 
if not then we can remove them

server/routes/reporting.js
```diff
@@ -1,11 +1,78 @@
...
+        bookingId,
+        originalLicence: res.locals.licence,
+        config: formConfig[formName],
+        userInput: req.body,
+        licenceSection: sectionName,
+        formName,
+        postRelease: res.locals.postRelease,
+      })
+
+      await standard.pushStatus(updatedLicence, formName, bookingId, req.user.username)
```
andrewrlee: presumably this never pushes to nomis?

This is derived by the config

server/routes/reporting.js
louisenorris: They do not, removed in latest commit

server/routes/reporting.js
louisenorris: Yep, `pushStatus` not needed and removed too, thanks!

andrewrlee (approved):
LGTM

PR #1715, opened 2026-06-04
Enforcing minimum npm version

andrewrlee:
Also installing npm-check-updates as a dev dependency - this fixes the set of dependencies that will be installed and means we can manage it explicitly


AledWEvans (approved):
LGTM

PR #1716, opened 2026-06-09
CVSL-3845 Initial GHA commit

andrewrlee:


gha-feature.env
```diff
@@ -0,0 +1,59 @@
+# true to enable test utils such as resetting DB
```
andrewrlee: all test vars - not real keys/secrets

.gitignore
```diff
@@ -27,6 +27,7 @@ licence-json/*
 out
 .gradle
 test-results/
```
andrewrlee, line 29: left the old one in as people will likely have old test results hanging around which we wouldn't want committing

dave-virden (approved):
LGTM

PR #1732, opened 2026-06-29
CVSL-4248 PROG HDC - VARY Early Adopter - Update HDC licence screen a…

AledWEvans:
…fter migration

### DIFF
--- a/bin/build-css
+++ b/bin/build-css
@@ -12,3 +12,5 @@
 ./node_modules/.bin/sass  "$@" \
     assets/sass/licences-pdf.scss \
     assets/stylesheets/licences-pdf.css
+./node_modules/.bin/sass  "$@" \
+    assets/stylesheets/task-hdc-to-cvl-notice.css
--- a/server/views/taskList/taskListTemplate.pug
+++ b/server/views/taskList/taskListTemplate.pug
@@ -9,13 +9,29 @@ include ./includes/viewButton
 include ./includes/viewEditButton
 include ./includes/utils
 
-
 block content
 
   if licenceVersion !== 0 && (user.role === 'RO' || user.role === 'CA')
     include ../includes/back-caselist-and-forms
   else
     include ../includes/backToCaseList
+
+
+  if (licenceInCvl)
+    div.govuk-inset-text.variationPanel
+      h2.heading-medium
+        span.icon-info i
+        span.title HDC licence variations
+      div.pure-g.pure-u-1.padLeft30
+        p If you need to vary this licence, you must now use the Create and vary a licence service.
+        p To do this, you must be either:
+
+        ul.midPaddingBottom
+          li the allocated community probation practitioner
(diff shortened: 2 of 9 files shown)


server/views/taskList/taskListTemplate.pug
```diff
@@ -9,14 +9,30 @@ include ./includes/viewButton
...
+        p If you need to vary this licence, you must now use the Create and vary a licence service.
+        p To do this, you must be either:
+
+        ul.midPaddingBottom
+          li the allocated community probation practitioner
+          li assigned to the same team as the allocated community probation practitioner on NDelius
+        p You will need to find the right person in the variation case list or search for them in Create and vary a licence.
+        p
+          a(href=createAndVaryALicenceVaryCaseloadUrl) Go to Create and vary a licence
+
```
AledWEvans, line 35: not testing licenceInCvl this as yet nothing seems to test this page in integration sense, might be a lot of work have tested in manually and passes fine?

bin/build-css
```diff
@@ -12,3 +12,5 @@
 ./node_modules/.bin/sass  "$@" \
     assets/sass/licences-pdf.scss \
     assets/stylesheets/licences-pdf.css
+./node_modules/.bin/sass  "$@" \
+    assets/stylesheets/task-hdc-to-cvl-notice.css
```
louisenorris, line 16: **question:** do we also need a file in `assets/sass/task-hdc-to-cvl-notice.scss` as is consistent with the added styling for `licences-pdf` and `forms-pdf`? Nor sure why we would but it follows previous implementation so asking the question.

louisenorris: Also suggesting we want to call it `migration-banner` or `variation-banner` rather than `task-hdc-to-cvl-notice`

AledWEvans: I think the licences-pdf.scss is in error or the css one, as they are duplicating the same css, as for the name yeha could be improved cvl-migration-banner-for-task-list? as we will create another banner for the case list!

server/views/taskList/taskListTemplate.pug
```diff
@@ -9,14 +9,30 @@ include ./includes/viewButton
...
 block content
 
   if licenceVersion !== 0 && (user.role === 'RO' || user.role === 'CA')
     include ../includes/back-caselist-and-forms
   else
     include ../includes/backToCaseList
 
+
+  if (licenceInCvl)
+    div.govuk-inset-text.variationPanel
```
louisenorris, line 21: if renaming following above suggestion then also suggest renaming to `variationPanel` to `variationBanner` but this is a nit pick comment 😅

bin/build-css
louisenorris: Cool, I couldn't work out why there was both, so just checking 😅

louisenorris (approved):
A couple of suggestive renaming comments but generally LGTM

PR #1739, opened 2026-07-13
CVSL-4275 add functionality to download migration log table as csv

gurps-singh:
This PR is to add functionality to download the migration log as a CSV. This now appears as an option on the migration log page.

Screenshot 2026-07-13 at 11 31 37


AledWEvans (approved):
LGTM

PR #1744, opened 2026-07-16
CVSL-4167 replace standard conditions in HDC

gurps-singh:
This PR is to replace the standard conditions in HDC following the  ISR Progression Model rollout. New licence conditions are being introduced so these need to be changed. 

Screenshot 2026-07-16 at 15 45 20

Screenshot 2026-07-16 at 15 46 16

Screenshot 2026-07-16 at 16 11 14


louisenorris (dismissed):
LGTM

AledWEvans (dismissed):
LGTM

server/views/licences/includes/licenceMixins.pug
```diff
@@ -101,28 +101,12 @@ mixin curfewConditions()
     | attached at the bottom of this licence.
 
 mixin standardConditions()
-  li Be of good behaviour and not behave in a way which undermines the purpose of the licence period;
```
andrewrlee, line 104: You can just leave these as part of the else block and remove `standardConditionsV2ForTemplate` - we want to preserve the existing formatting and text on the current licence

andrewrlee (approved):
LGTM

PR #1747, opened 2026-07-20
CVL | HDC optimise call to reduce unnecessary API calls

AledWEvans:
small check to prevent calls when not needed


andrewrlee (approved):
LGTM

PR #1748, opened 2026-07-22
CVSL-4373 update types, drop com email from licence migration log table and display prison number on licence migration log

louisenorris:


andrewrlee (approved):
LGTM

PR #1768, opened 2026-07-29
CVSL-4346 update devsec hooks

andrewrlee:
This may require installing gitleaks if you don't already have it installed


nishanthmahasamudram (approved):
LGTM

